PR #1, opened 2014-05-26
Fixed path with php_extension_conf_path var in Debian

dcastello:
The path to php_extension_conf_path variable in Debian systems (Ubuntu) is wrong.
The right path is "/etc/php5/apache2/conf.d" instead of "/etc/php5/conf.d".


geerlingguy:
Closing this; I'm going to be fixing extension/config paths for Debian and RedHat once and for all in a bit... finally had to get everything working on Debian, RedHat, CentOS, and Ubuntu all in one day, and it's such a pain!

PR #21, opened 2014-11-20
Avoid expired caches

M1r1k:
Run apt-get update before fetching php packages.


geerlingguy:
Thanks! I've added it in a format similar to other roles, so it doesn't necessarily update the apt caches on every single occurrence of the task.

PR #27, opened 2015-03-25
Changes

drupaladmin:


geerlingguy:
@drupaladmin - Since this role is used by a large number of people who might be depending on portions of the functionality current provided (e.g. `php-fpm` configuration out of the box), and who also often don't use MySQL (so it shouldn't be a default), could you open up separate pull requests per bit of functionality? That way I could more easily see which ones are flexible/stable enough for quick acceptance, and which ones would need more work/discussion.

Issue #24, opened 2015-02-28
Configure opcache for PHP >= 5.5


geerlingguy:
Opcache is the new APC. Allow it to be configured if installed version of PHP is >= 5.5.

geerlingguy:
See also, #28

geerlingguy:
And to highlight the importance of using the OpCache...

```
# OpCache disabled, PHP 5.5.23, Apache 2, Drupal 8 HEAD
$ wrk -t1 -c100 -d10 http://drupaltest.dev/
Running 10s test @ http://drupaltest.dev/
  1 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   908.50ms  532.07ms   1.99s    66.18%
    Req/Sec    35.07     33.38   138.00     78.18%
  231 requests in 10.06s, 2.18MB read
  Socket errors: connect 0, read 0, write 0, timeout 163
Requests/sec:     22.96
Transfer/sec:    221.88KB

# OpCache enabled, PHP 5.5.23, Apache 2, Drupal 8 HEAD
$ wrk -t1 -c100 -d10 http://drupaltest.dev/
Running 10s test @ http://drupaltest.dev/
  1 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency   163.89ms  223.99ms   1.88s    88.29%
    Req/Sec   220.67     90.81   380.00     62.63%
  2195 requests in 10.06s, 20.89MB read
  Socket errors: connect 0, read 0, write 0, timeout 79
Requests/sec:    218.29
Transfer/sec:      2.08MB

PR #28, opened 2015-03-29
Enable basic OpCache configuration.

mglaman:
Can't install Drupal Commons when using this role because OpCache requirement is 128 and no way to customize.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -16,6 +16,11 @@ php_apc_shm_size: "96M"
 php_apc_stat: "1"
 php_apc_enable_cli: "0"
 
+# OpCache settings
+php_opcache_enable: 1
+php_opcache_enable_cli: 0
+php_opcache_memory_consumption: 64
+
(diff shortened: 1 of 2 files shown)


geerlingguy:
@mglaman - A few notes/questions:

  - Will these changes still work with older versions of PHP, or could they cause issues? (e.g. PHP 5.3/5.4).
  - Can you add the new defaults to the README as well, and explain what the settings do (similar to the APC section)?
  - Why all the extra commented vars? We should probably leave them out, since users can look in the PHP docs for default values.

geerlingguy:
@mglaman - I've been working on a more comprehensive addition that works with PHP 5.3+, on RHEL + Ubuntu, similar to how the APC integration works, and I'll be completing that work over in #24. Luckily, it uses the _exact same_ variable names that you added in this PR, so you shouldn't need to update your downstream playbooks :)

mglaman:
@geerlingguy sounds good :)

PR #29, opened 2015-04-12
add configurable max post size

louim:
I know there is another pull request for this (#27), but this one applies cleanly to the master and has the default already set.


geerlingguy:
Thanks!

PR #31, opened 2015-04-14
Add option to disable expose_php

louim:
I think it's an important setting, specially for security. Since I needed it, I made a pull request at the same time.


stevenscg:
+1. Thanks @louim. I just implemented your same changes.

PR #34, opened 2015-04-22
Update README.md

kienpham2000:
my ubuntu can't seem to find old php packages so updated to php5- prefix.


geerlingguy:
This should be okay now; the default Debian/Ubuntu package list is picked up from the `vars/Debian.yml` file, and the README only reflects one set of packages, simply for convenience's sake.

PR #36, opened 2015-04-24
Cli environment configuration

Gta-Cool:
Adding a configured php.ini file for the CLI environment.

By default, the PHP CLI environment uses the same variables as the web server environment, but they can be redefined by CLI specific variables.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,3 +35,10 @@ script:
     | grep -q 'The PHP Group'
     && (echo 'PHP is installed' && exit 0)
     || (echo 'PHP is not installed' && exit 1)
+
+  # Make sure PHP cli is correctly configured.
+  - >
+    php -r "echo ini_get('date.timezone'), \"\\n\";"
+    | grep -q 'Europe/London'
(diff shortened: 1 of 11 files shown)


guillaumecodet:
:+1:

oktapodia:
:+1:

geerlingguy:
I'm fine with adding a separate configuration for CLI, but the extra logic inside the php.ini.j2 template is a bit jarring. I'd prefer if we split it into a php.ini.j2 and php-cli.ini.j2 instead, even though that creates two templates.

geerlingguy:
After giving this some more thought, I'd rather not support a separate CLI ini in this role. The role has become complex enough in its current state and is focused primarily on use cases where web/FPM are the primary uses of this role's configuration, and CLI is secondary.

I would rather recommend someone to a separate `php-cli` role that just configures CLI separately after this role completes general PHP setup. If there were a couple simple changes to _this_ role to make it so that kind of setup would maintain idempotence and work well, I'd be willing to make those, but this adds a lot of overhead for functionality that I'm certain the majority of this role's users wouldn't require/modify :(

PR #43, opened 2015-06-06
PHP-FPM configuration

JoelGilliland:
Added php-fpm configuration options.

I don't care for some of the logic in the www.conf.j2 template as it stands, but it's something I wanted/needed myself. Debian is currently untested. Open to suggestions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,18 +28,6 @@ The default values for the HTTP server deamon are `httpd` (used by Apache) for R
 
 (RedHat/CentOS only) If you have enabled any additional repositories (might I suggest geerlingguy.repo-epel or geerlingguy.repo-remi), those repositories can be listed under this variable (e.g. `remi,epel`). This can be handy, as an example, if you want to install the latest version of PHP 5.4, which is in the Remi repository.
 
-### PHP-FPM
-
-PHP-FPM is a simple and robust FastCGI Process Manager for PHP. It can dramatically ease scaling of PHP apps and is the normal way of running PHP-based sites and apps when using a webserver like Nginx (though it can be used with other webservers just as easily).
-
-When using this role with PHP running as `php-fpm` instead of as a process inside a webserver (e.g. Apache's `mod_php`), you need to set the following variable to `true`:
-
-    php_enable_php_fpm: false
-
-You will also need to override the default `php_packages` list and add `php-fpm` (RedHat/CentOS) or `php5-fpm` (Debian/Ubuntu) to the list.
-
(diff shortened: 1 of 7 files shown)


geerlingguy:
I like what you're doing here, and I think I'll take the role this route—currently I manage all the FPM configuration for my prod servers outside of this role, but it'd be nice if this role can handle the configuration as well.

Over in #45 I added some initial/minimal configuration that is only applied when building from source. I'd like to generalize it so it works on both Debian/Ubuntu _and_ CentOS/RHEL (my work is currently only tested on Debian).

sagikazarmark:
:+1: Supporting multiple pools would be nice.

sagikazarmark:
This role has quite good pool handling: https://github.com/NBZ4live/ansible-php-fpm

etsglobalbv:
Thanks @JoelGilliland for this issue :
@geerlingguy can you merge it ?

It is very usefull when you build a nginx-php with vagrant as you need to change the user and group to "vagrant" instead of www-data

bpiselli:
Thanks @JoelGilliland for this issue :
@geerlingguy can you merge it ?

It is very usefull when you build a nginx-php with vagrant as you need to change the user and group to "vagrant" instead of www-data

JoelGilliland:
@sagikazarmark I haven't need multiple pools yet, but there is a good chance I might be needing it in the very near future. I'll try to get it added to my fork soon.

bpiselli:
Thanks for the fast answer @JoelGilliland

geerlingguy:
FYI I see you point about allowing the user override, but my main goal with this role (and most of my others) is to have a base configuration that works out of the box no matter what situation you have (e.g. Apache or Nginx, standalone, etc.).

If we add a default, then we either need more complicated logic to determine the default user to set (e.g. my crazy-complicated Apache role that covers two versions of Apache on four different OSes, each with different pathing requirements for config!), or we make it so certain users _always_ have to add an override, which isn't the end of the world of course, but I'd prefer it to work straight out of the box.

Instead, I'm taking an approach of only adjusting the `listen` and basic `pm.*` variables out of the box with the above commit. These are the values that I have to adjust 80% of the time (I rarely change the user from the default of www-data (or apache if using with Apache), and if you want to change that while using this role, you can add a task in your playbook like:

```yaml
- name: Configure php-fpm pool (if enabled).
  lineinfile:
    dest: "{{ php_fpm_pool_conf_path }}"
    regexp: "{{ item.regexp }}"
    line: "{{ item.line }}"
    state: present
  with_items:
    - regexp: "^user.?=.+$"
      line: "user = vagrant"
    - regexp: '^group.?=.+$'
      line: "group = vagrant"
  when: php_enable_php_fpm
  notify: restart php-fpm
```

I'll consider adding more vars over time, but each additional variable/line of configuration needs to be justified as being something 80% of this roles users would benefit from. One of my main goals with this role is to try to keep it maintainable/grokkable by most devs, and the further I go with slightly-more-advanced features like pool management, the further away from the 80% use case I get.

For more flexibility, I'd recommend either using this role then adding in extra tasks in your playbook like the above, forking this role and adding what you need, or using a different role that is more geared toward a specific use case you need.

PR #54, opened 2015-09-12
Variable to enable / disable php_allow_url_fopen

csedev:
According to phpsec.org (http://phpsec.org/projects/phpsecinfo/tests/allow_url_fopen.html), php_allow_url_fopen should be disabled for security reasons.

It's also recommended to be disabled for Drupal installations (https://www.drupal.org/requirements/php).

I've created a variable "php_allow_url_fopen" and set the default to "Off". This allows the user to override the value if they desire.

Builds successfully using Travis CI (https://travis-ci.org/csedev/ansible-role-php/jobs/80001558)

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -45,6 +45,8 @@ php_upload_max_filesize: "64M"
 php_post_max_size: "32M"
 php_date_timezone: "America/Chicago"
 
+php_allow_url_fopen: "Off"
+
 php_sendmail_path: "/usr/sbin/sendmail -t -i"
 php_short_open_tag: false
 
--- a/templates/php.ini.j2
+++ b/templates/php.ini.j2
@@ -88,7 +88,7 @@ max_file_uploads = 20
 ; Fopen wrappers ;
 ;;;;;;;;;;;;;;;;;;
 
-allow_url_fopen = On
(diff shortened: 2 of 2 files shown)


defaults/main.yml
```diff
@@ -45,6 +45,8 @@ php_upload_max_filesize: "64M"
 php_post_max_size: "32M"
 php_date_timezone: "America/Chicago"
 
+php_allow_url_fopen: "Off"
+
```
geerlingguy: Can you remove this extra line? This can be grouped with other 'misc options' instead of being on its own.

defaults/main.yml
```diff
@@ -45,6 +45,8 @@ php_upload_max_filesize: "64M"
 php_post_max_size: "32M"
 php_date_timezone: "America/Chicago"
 
+php_allow_url_fopen: "Off"
```
geerlingguy: Unfortunately, because this defaults to 'On' in PHP currently, and because there are dozens of apps/frameworks that are already using the functionality (instead of curl, Guzzle, or other solutions), I'd rather default this to "On" so I don't break all those apps with the next role version bump.

There _are_ valid use cases for setting this to On, though as phpsec and you point out, it should be pretty rare, and for most people, it's so much safer to _not_ leave it On :)

PR #55, opened 2015-09-14
php-fpm extra configuration params

alfonsodev:
This pull request adds some flexibility in the php-fpm configuration.


geerlingguy:
@alfonsodev - Thanks for the PR! I've gone down a slightly different path after considering a few other options, and here are some of my reasons: https://github.com/geerlingguy/ansible-role-php/pull/43#issuecomment-171853248

PR #58, opened 2015-11-09
Support modifiyng session.save_handler and session.save_path using variables

obokaman-com:
We're using Redis to manage PHP sessions, so I've find the need to define session.save_handler and session.save_path in order to do this. I think it could be useful to include as an available configuration variable.


geerlingguy:
Tried a different approach—the save path doesn't need to be in a conditional, that seems to have been throwing off the tests.

PR #60, opened 2015-12-02
configure php.ini for fpm

teadur:
configure php.ini for fpm if fpm is enabled and ini is managed.

### DIFF
--- a/tasks/configure.yml
+++ b/tasks/configure.yml
@@ -46,3 +46,16 @@
     state: started
     enabled: yes
   when: php_enable_php_fpm
+
+- name: Place PHP configuration file for fpm in place.
+  template:
+    src: php.ini.j2
+    dest: "{{ php_conf_path }}/fpm/php.ini"
+    owner: root
+    group: root
+    mode: 0644
+  notify: restart php-fpm
+  when: php_use_managed_ini
+  when: php_enable_php_fpm
+
+
--- a/tasks/install-from-source.yml
+++ b/tasks/install-from-source.yml
@@ -39,6 +39,7 @@
     - libmysqlclient-dev
     - libpspell-dev
     - librecode-dev
+    - libssl-dev
   when: ansible_os_family == 'Debian'
 
 - name: Check if gmp.h is already in a location accessible to gcc.


geerlingguy:
One quick question—is the `fpm` path the same (relative to the configuration path) across RedHat and Debian flavors? Otherwise we'll need to somehow switch for that between the two. Might be different.

teadur:
not sure about rhel, im using debian jessie for my deployment.

tasks/configure.yml
```diff
@@ -51,3 +51,16 @@
...
+
+- name: Place PHP configuration file for fpm in place.
+  template:
+    src: php.ini.j2
+    dest: "{{ php_conf_path }}/fpm/php.ini"
+    owner: root
+    group: root
+    mode: 0644
+  notify: restart php-fpm
+  when: php_use_managed_ini
```
geerlingguy: These two 'when's need to be combined (e.g. `php_use_managed_ini and php_enable_php_fpm`) else only the last defined one will be used.

Draiken:
Shouldn't this just be defined on the vars folder?

[Current logic](https://github.com/geerlingguy/ansible-role-php/blob/master/vars/Debian.yml#L17) only checks if it's configured as a daemon, doesn't check if fpm is used at all.
if we just add the check if fpm is enabled and set the conf path (on Debian) to `/etc/php/fpm` I imagine everything should work as expected.

If we can allow both daemon and FPM, this needs to be a specific step like proposed here. But I'd put this step on configure-fpm

tasks/configure.yml
```diff
@@ -51,3 +51,16 @@
...
+  when: php_enable_php_fpm
+
+
```
geerlingguy: Needs one fewer line break here.

oxyc:
So I guess these are the paths:

Ubuntu 12.04 and 14.04:
- no webserver: `/etc/php5/php.ini`
- nginx: `/etc/php5/php.ini`
- apache fpm: `/etc/php5/fpm/php.ini` (this scenario is broken with this commit)
- apache mod_php: `/etc/php5/apache/php.ini`
- cgi: `/etc/php5/cgi/php.ini`
- etc...

CentOS:
- no webserver: `/etc/php.ini`
- nginx: `/etc/php.ini`
- apache fpm: `/etc/php.ini`
- apache mod_php: `/etc/php.ini`
- cgi: `/etc/php.ini`
- etc...

What about simply doing like you did with xhprof https://github.com/geerlingguy/ansible-role-php-xhprof/commit/937322d3786ad83d6ce897516feb759f20e935cd?

geerlingguy:
@oxyc - I'd be amenable to either solution. And see why I love CentOS/RHEL so much? There's _consistency_ :P

geerlingguy:
This was fixed by #83 - thanks everyone for the work on this!

PR #61, opened 2015-12-03
Enable php5-fpm in apache2 configuration

teunis90:
php5-fpm can be installed from binaries nowadays, so I've added the possibility to activate the config in apache via this role. I've chosen to put the configuration in the modules directory for backwards compatibility since Debian changed the conf.d directory to conf-available / conf-enabled. 

The change is only for Debian based systems and is tested with the following variables:

    # Php
    php_packages:
    - php5-fpm
    php_enable_webserver: True
    php_enable_php_fpm: True

### DIFF
--- a/tasks/configure-Debian.yml
+++ b/tasks/configure-Debian.yml
@@ -1,0 +1,12 @@
+- name: Ensure php5-fpm config file is installed.
+  template:
+    src: apache2-fpm.j2
+    dest: /etc/apache2/mods-available/php5-fpm.conf
+    mode: 0644
+  when: php_enable_php_fpm and php_webserver_daemon == "apache2" 
+  notify: restart webserver
+
+- name: Ensure php5-fpm config file is activated
+  file: src=../mods-available/php5-fpm.conf dest=/etc/apache2/mods-enabled/php5-fpm.conf state=link  
+  when: php_enable_php_fpm and php_webserver_daemon == "apache2" 
+  notify: restart webserver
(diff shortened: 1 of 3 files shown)


geerlingguy:
For Debian, the Apache2 2.4 setup already uses FPM by default, IIRC. Is this not correct?

teunis90:
Are you sure? I did not found any default package file which configures the connection to the FPM daemon.

geerlingguy:
@teunis90 - You're correct, I must be remembering some incorrect information.

teunis90:
@geerlingguy Do you like to have this part of the configuration in geerlingguy.apache or geerlingguy.php role? We use both so would be nice to stay in sync.

geerlingguy:
@teunis90 - The actual Apache config (where it's adding a config to Apache and requires an Apache restart) should be in the `apache` role, and maybe be triggered by a variable like `apache_use_php_fpm` with true/false.

geerlingguy:
@teunis90 - I've had some more time to play around with different options, and to keep things a little simpler, I decided to add a new, separate role, [geerlingguy.apache-php-fpm](https://github.com/geerlingguy/ansible-role-apache-php-fpm), which connects Apache to PHP-FPM.

To use it with Apache, you simply set a line in your virtualhost to direct requests for .php files to FPM:

    ProxyPassMatch ^/(.*\.php(/.*)?)$ "fcgi://127.0.0.1:9000/var/www/example"

Or if you're using a socket for `listen` instead of a TCP connection:

    ProxyPassMatch ^/(.*\.php(/.*)?)$ "unix:/var/run/php5-fpm.sock|fcgi://localhost/var/www/example"

PR #62, opened 2015-12-07
Allow overriding php_fpm_daemon

thom8:
Allow overriding `php_fpm_daemon` in a similar way to other vars.


geerlingguy:
Fail was unrelated to this patch. Looks great!

PR #66, opened 2015-12-14
Use ansible jinja filter of version_compare

pgilad:


geerlingguy:
TIL. I'm amazed I haven't noticed that filter before. Thanks for the PR, I'll be using this on my apache role too.

PR #68, opened 2016-01-07
Improve repo description

remicollet:
* add link to referenced roles
* use remi-php56 as example (instead of EOL php 5.4)


geerlingguy:
Thanks! This does improve the docs a bit, and I'll roll it into the next tagged release of the role with some other changes.

PR #72, opened 2016-01-22
Made php_fpm_pool_conf_path configurable - for PHP 7.

LewisW:
Unfortunately Debian likes to have everything PHP5 related in a PHP5 directory, so in-order to get this role working with PHP 7, yet another path has to be configurable!


geerlingguy:
Debian--

PR #78, opened 2016-02-10
Fix php_fpm_listen for Debian OS

opdavies:
Update the value of `php_fpm_listen` for Debian/Ubuntu servers.

### DIFF
--- a/vars/Debian.yml
+++ b/vars/Debian.yml
@@ -21,3 +21,5 @@ __php_apc_conf_filename: 20-apcu.ini
 __php_opcache_conf_filename: 05-opcache.ini
 __php_fpm_daemon: php5-fpm
 __php_fpm_pool_conf_path: "/etc/php5/fpm/pool.d/www.conf"
+
+php_fpm_listen: "unix://var/run/php5-fpm.sock"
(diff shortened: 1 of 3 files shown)


vars/Debian.yml
```diff
@@ -21,3 +21,5 @@ __php_apc_conf_filename: 20-apcu.ini
 __php_opcache_conf_filename: 05-opcache.ini
 __php_fpm_daemon: php5-fpm
 __php_fpm_pool_conf_path: "/etc/php5/fpm/pool.d/www.conf"
+
+php_fpm_listen: "unix://var/run/php5-fpm.sock"
```
geerlingguy: I would like to keep this variable the same for all OSes, and keep the default as the TCP/IP port.

I'm not entirely sure what this would give us—and the variable would no longer be easy to override upstream if we put it in an OS-specific included vars file.

opdavies: With a fresh Drupal VM install with `geerlingguy/ubuntu1404` using Nginx, the box provisions correctly but Nginx doesn't render the page - instead it returns a 502 Bad Gateway error.

I've managed to get around it for now by adding `php_fpm_listen: "unix://var/run/php5-fpm.sock"` into Drupal VM's config.yml and re-provisioning.

geerlingguy:
I'm going to make sure this is fixed upstream.

Issue #64, opened 2015-12-14
Use Docker-based Travis CI tests for multi-platform support


geerlingguy:
See similar: https://github.com/geerlingguy/ansible-role-apache/pull/60

geerlingguy:
One major difference that makes this a slightly more complex role... I currently have two builds—install from package and install from source. And I would need to do both on each OS to be complete:

```
env:
  - SITE=test-install-package.yml
  - SITE=test-install-from-source.yml
```

So... trying to figure out the simplest way to do a build matrix without slowing down the build and causing annoying code duplication.

geerlingguy:
This will help a _ton_ with the insane complexity of multi-OS/multi-version configuration support—currently PHP's main config can be located in a total of four separate locations on Debian and at least two on CentOS, and those locations also change depending on whether you're using Apache or Nginx, and what version of Debian/CentOS you're using.

I hate making changes or accepting PRs that break one of the four platforms, or do things like miss config for FPM vs. CLI, or Apache2 vs FPM, etc... and this issue will help catch those issues before they _become_ issues. So I'm going to kind of put the rest of the role improvements on hold (even though I'm dying to get FPM config better here!) until I get this done.

geerlingguy:
This was fixed in https://github.com/geerlingguy/ansible-role-php/pull/80

PR #80, opened 2016-02-16
Issue #64: Add docker based tests

oxyc:


geerlingguy:
One annoying thing—I can't for the life of me figure out the syntax for `allowed_failures` in the travis build matrix; I would like to still have CentOS 6/7 tests run in some situations even though I know there's a failure I'd have to deal with, because it's useful to see if PRs break anything _unexpected_. If you can't get tests to pass on non-Ubuntu platforms, then if you know any tricks to setting up the matrix for allowed failures that would also be useful as a stopgap...

oxyc:
Should be fairly easy to get FPM tests going (lets see how this works). But testing `mod_php` sounds like a nightmare across all OSes.

oxyc:
@geerlingguy any ideas on how to test this for all variations? Spinning up 3*4 travis instances is a bit too much as well...

geerlingguy:
@oxyc - Could just take the approach of testing the defaults on all environments, then testing different configs on only CentOS 7 and Ubuntu 14.04... I don't want to kill Travis with a perfect build matrix.

oxyc:
@geerlingguy yes I copied from the git role and I think all builds pass except for centos 7 from source at the moment (working on it in the last commit).

Would be nice to test fpm though, so we can answer your question here https://github.com/geerlingguy/ansible-role-php/pull/60#issuecomment-162689342

oxyc:
Maybe not worth it like you say. On #60 we could just test it manually and make sure it works before committing.

I'll stick to testing packages and source in this commit (like in the git role).

geerlingguy:
Manual test is fine by me.

oxyc:
CentOS 6 fails on this: https://travis-ci.org/geerlingguy/ansible-role-php/jobs/109643689#L866

Should we support the 5.3 package on CentOS 6 or have a higher version as a dependency?

geerlingguy:
@oxyc - Let's make higher version a dep - we should test with at least 5.5, maybe even just do 5.6 by default. If people want to use older (essentially dead) versions from the system repo, then it's up to them to make sure their config works correctly.

oxyc:
It seems re2c is not in the base yum repo :/

oxyc:
@geerlingguy I'll see how far I can get the source build on centos6 as `re2c` is available there. I do not dare adding random `rpm`s, so maybe you can have a look at that in a follow up task.

geerlingguy:
K.

oxyc:
There's this one http://pkgs.org/centos-7/russian-fedora-free-x86_64/re2c-0.13.5-7.el7.R.x86_64.rpm.html. Alternatively building it from source.

geerlingguy:
Haha, the tangled web we weave...

geerlingguy:
We could add a shell script to run inside the `tests` folder since this is really just for tests, though, right?

oxyc:
As `re2c` isnt available for RHEL7, I can't see how the build from source option works for CentOS 7. Which is something that's supposed to be supported right?

oxyc:
Okay maybe it's not actually necessary (I saw it listed as a dependency in php-src INSTALL notes). I'll setup docker locally so I can test this easier.

Sorry for all the confusion, I'll post back here tomorrow once I have done some proper debugging.

geerlingguy:
@oxyc - Yes, the source build was working in 7 last time I tested locally... but that was a while back. Might be something docker specific. Also, sorry I haven't been able to help much more today—this week I'm wrapping a large project, but next week I can put more time into this :)

geerlingguy:
See related: https://bugzilla.redhat.com/show_bug.cgi?id=1212912

oxyc:
No worries, I imagine you have enough roles on your hands already :smile:

I actually got the build passing on my CentOS 7 VM, so I probably just need to get the dependencies correct. The reason I thought `re2c` was a hard dependency was this scary looking ansible output https://travis-ci.org/geerlingguy/ansible-role-php/jobs/109746150#L895. Then I noticed it was just a warning though...

oxyc:
Yey, got it to pass. So the warning wasn't anything to worry about.

geerlingguy:
Woohoo! Checking it over, then will merge and we'll be on our merry way towards stabilizing this role.

Issue #74, opened 2016-01-29
Set always_populate_raw_post_data to -1 by default


mikebell:
On a default provision you can't enable the REST module in D8 because ```always_populate_raw_post_data``` should be set to ```-1``` in php.ini

It'd be nice if this was set by default.

geerlingguy:
Hmm... this configuration variable is removed in 7.0.0, so I might not want to add it in (see: http://php.net/manual/en/ini.core.php). Is there any other way you can set that var at runtime, in settings.php, or elsewhere?

geerlingguy:
This was fixed in #81

PR #81, opened 2016-02-19
Set always_populate_raw_post_data to -1 as setting is deprecated in 5.6

wshafer:
More info: http://php.net/manual/en/ini.core.php#ini.always-populate-raw-post-data

### DIFF
--- a/templates/php.ini.j2
+++ b/templates/php.ini.j2
@@ -64,6 +64,7 @@ auto_prepend_file =
 auto_append_file =
 
 default_mimetype = "text/html"
+always_populate_raw_post_data = -1
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;
 ; Paths and Directories ;


geerlingguy:
If we set this, will PHP 7.0 break from having a variable defined that's been removed? This role should work across PHP 5.6 and PHP 7+.

See also: https://github.com/geerlingguy/ansible-role-php/issues/74

wshafer:
Well if you want that to work with 5.6 AND 7 the best is to set this val to -1.  PHP 7 should just ignore this setting, but if not set to -1 PHP 5.6 throws deprecation errors on every request.

wshafer:
Just spun up a PHP 7 to test.  This setting is ignored.

```bash
vagrant@php7dev:~$ php -v
PHP 7.0.5-dev (cli) (built: Feb 19 2016 16:54:38) ( NTS DEBUG )
Copyright (c) 1997-2016 The PHP Group
Zend Engine v3.0.0, Copyright (c) 1998-2016 Zend Technologies
    with Zend OPcache v7.0.6-dev, Copyright (c) 1999-2016, by Zend Technologies
vagrant@php7dev:~$ php --ini
Configuration File (php.ini) Path: /etc/php7
Loaded Configuration File:         /etc/php7/php-cli.ini
Scan for additional .ini files in: /etc/php7/conf.d
Additional .ini files parsed:      (none)
vagrant@php7dev:~$ grep "always_populate_raw_post_data" /etc/php7/php-cli.ini
 always_populate_raw_post_data = -1
vagrant@php7dev:~$ grep "error_reporting" /etc/php7/php-cli.ini
error_reporting=E_ALL
vagrant@php7dev:~$ php -r 'echo "hello there\n";'
hello there
```

geerlingguy:
Perfect! I'll merge this then.

wshafer:
Awesome.  BTW.. Thanks for putting all this together.  Both this rule and nginx have been working perfectly.

PR #82, opened 2016-02-20
Test them all

oxyc:
So something messed up my VM where I was testing this locally... Takes too long to download so I'm just using travis instead. Tests are temporary and you cant actually look at the travis result. Just look at the output of the `curl .. | grep 'memory_limit.*191M'` test

Apache with mod_php
- [x] Centos 6
- [x] Centos 7
- [x] Ubuntu 12.04
- [x] Ubuntu 14.04

Apache with FPM
- [ ] Centos 6
- [x] Centos 7
- [ ] Ubuntu 12.04
- [x] Ubuntu 14.04

~~Nginx with regular PHP~~
- ~~[ ] Centos 6~~
- ~~[ ] Centos 7~~
- ~~[ ] Ubuntu 12.04~~
- ~~[ ] Ubuntu 14.04~~

Nginx with FPM
- [x] Centos 6
- [x] Centos 7
- [ ] Ubuntu 12.04
- [x] Ubuntu 14.04

No webserver
- [x] Centos 6
- [x] Centos 7
- [x] Ubuntu 12.04
- [x] Ubuntu 14.04


oxyc:
Hmm so what happens if you run `nginx` without PHP FPM?

oxyc:
@geerlingguy I suspect there's something incorrect about how I've set up the nginx hosts. Any ideas?

oxyc:
One issue with this is that we dont support cgi on Ubuntu. I believe you use it in some of your examples so maybe we should add a task for detecting if `libapache2-mod-fastcgi` is installed?

oxyc:
I'll close this PR but keep it for historical reference if we ever need to run test in various scenarios again.

See #83

PR #83, opened 2016-02-20
Closes #53, #60: Fix php.ini not applied for php-fpm on ubuntu

### LINKED ISSUE #53: php.ini is not applied in certain situations on Ubuntu
Hello, I'm not sure whether it's a bug.
Role creates a configuration in /etc/php5/php.ini, but PHP use another files:
/etc/php5/cli/php.ini - for cli, and
/etc/php5/fpm/php.ini - for php-fpm

This forces to write additional code
```
    - name: Copy php.ini to php5/cli
      command: cp /etc/php5/php.ini /etc/php5/cli/php.ini

    - name: Copy php.ini to php5/fpm
      command: cp /etc/php5/php.ini /etc/php5/fpm/php.ini
      notify: restart php-fpm
```

oxyc:
This is a mergable PR for #82. I want to keep the old PR for historical reference in case we need to do similar scenario testing again. This works for all distros:
- Apache with `mod_php`
- Apache with PHP FPM (not on Centos 6 because the test setup is complex, this bug fix doesnt adress centos anyway)
- Nginx with PHP FPM
- No webserver (applies to the CLI php).

I feel this is the best approach as a configuration change and a re-provision would create the new ini only if necessary. Instead of bloating the configs or complicating the role unnecessarily (trying to create apc/opcache ini files for all directories would be a nasty `with_nested` task). 

The docker build for PHP FPM only pass on Ubuntu 14.04 with the custom initctl mentioned [here](https://github.com/geerlingguy/ansible-role-mysql/pull/84#issuecomment-186270022). Ubuntu 12.04 fails in the same way as in the before mentioned task. But the ini directories are the same for both 12.04 and 14.04.

CentoOS 6 tests for Apache with PHP FPM fail because of https://github.com/geerlingguy/drupal-vm/issues/346#issuecomment-171869866

The only questionable addition would be to somehow support `libapache2-mod-fastcgi`. This would mean detecting if the package is installed and if so, creating the `php.ini` in `/etc/php5/cgi/php.ini` instead.

### DIFF
--- a/vars/Debian.yml
+++ b/vars/Debian.yml
@@ -14,7 +14,7 @@ __php_packages:
 __php_webserver_daemon: "apache2"
 
 # Vendor-specific configuration paths on Debian/Ubuntu make my brain asplode.
-__php_conf_path: "{{ '/etc/php5' if php_webserver_daemon and php_webserver_daemon != 'apache2' else '/etc/php5/apache2' }}"
+__php_conf_path: "{{ '/etc/php5/fpm' if php_enable_webserver and php_enable_php_fpm else '/etc/php5/apache2' if php_enable_webserver and php_webserver_daemon == 'apache2' else '/etc/php5/cli' }}"
 __php_extension_conf_path: "{{ __php_conf_path }}/conf.d"
 
 __php_apc_conf_filename: 20-apcu.ini


geerlingguy:
Is the idempotence failure something having to do with the `service` task?

And thanks _so much_ for this work so far—making my weekend and the next Drupal VM much easier!

oxyc:
Hmm that's odd... let's see what's causing it.

geerlingguy:
It looks like the two tasks that break idempotency are:

```
TASK [geerlingguy.php : Place APC configuration file in place.] ****************
changed: [ubuntu1404]

TASK [geerlingguy.php : Place OpCache configuration file in place.] ************
changed: [ubuntu1404]
```

(Doing some local testing on VMWare).

After the second run, everything's good to go :/

oxyc:
Hmm very strange O.o I don't see how this commit good have caused that. Unfortunately I wont be able to test this more today. Hope you can figure out what's going on.

geerlingguy:
Found it - Debian/Ubuntu configuration insanity strikes again!

Here's what the config directory looks like immediately after PHP's first installation/fpm starts up:

```
vagrant@ubuntu1404test:/etc/php5/cli/conf.d$ ls -lah
total 8.0K
drwxr-xr-x 2 root root 4.0K Feb 20 22:24 .
drwxr-xr-x 3 root root 4.0K Feb 20 22:24 ..
lrwxrwxrwx 1 root root   32 Feb 20 22:24 05-opcache.ini -> ../../mods-available/opcache.ini
lrwxrwxrwx 1 root root   28 Feb 20 22:24 10-pdo.ini -> ../../mods-available/pdo.ini
lrwxrwxrwx 1 root root   29 Feb 20 22:24 20-apcu.ini -> ../../mods-available/apcu.ini
lrwxrwxrwx 1 root root   29 Feb 20 22:24 20-curl.ini -> ../../mods-available/curl.ini
lrwxrwxrwx 1 root root   27 Feb 20 22:24 20-gd.ini -> ../../mods-available/gd.ini
lrwxrwxrwx 1 root root   29 Feb 20 22:24 20-json.ini -> ../../mods-available/json.ini
lrwxrwxrwx 1 root root   31 Feb 20 22:24 20-mcrypt.ini -> ../../mods-available/mcrypt.ini
lrwxrwxrwx 1 root root   33 Feb 20 22:24 20-readline.ini -> ../../mods-available/readline.ini
```

Note that those are all symlinks to files inside `mods-available`.

After the first provision, that's still the structure of that directory. But after the second provision, the templated files _are_ copied into place (since we have `force: yes`).

I'm going to fiddle with the ordering.

oxyc:
Ugh. I do see what you mean about Ubuntu... I think I'll begin using centos for my base boxes after all this :smile:

geerlingguy:
Yay! I have a fix.

geerlingguy:
Adding more tasks for Debian because it's weird--

geerlingguy:
We'll see if the fix works; tested 3 times with different settings in 12/14 locally.

PR #85, opened 2016-02-29
Missing dependencies for installing PHP from source on Ubuntu 15.10

oxyc:


oxyc:
We could also specify the openssl directory but simply requiring `pkg-config` sounds better imo. https://github.com/mongodb/mongo-php-driver/issues/138#issuecomment-156774086

oxyc:
This is mergable. Here's a build which passed on Ubuntu 15.10 https://travis-ci.org/geerlingguy/ansible-role-php/builds/112639945

geerlingguy:
Looks like `pkg-config` is in `trusty` and `precise`, so I'm fine with it.

Issue #48, opened 2015-07-27
opcache_enable doesn't change 


MauroNigrele:
Hi, if you have php_opcache_enable = '1' and then set to '0' change have no effect since configuration task depends on this value. 

@see: https://github.com/geerlingguy/ansible-role-php/blob/master/tasks/configure.yml#L40

I can do a PR solving this, but I don't know if removing this line could do some side-effect

MauroNigrele:
I found that the configuration file:  _/etc/php5/conf.d/05-opcache.ini_ isn't used by **php-fpm** which is using _/etc/php5/fpm/conf.d/05-opcache.ini_ (link of **_/etc/php5/mods-available/opcache.ini**) instead, so maybe the problem is more complex and we would need to add some new var and task to configure this file too.

geerlingguy:
You can submit a PR—just removing that line should be okay, and I don't think it should cause any issue, since `php_opcache_enabled_in_ini` is `false` by default, so it won't try enabling anything that isn't enable-able.

MauroNigrele:
Ok Jeff, i will try to solve the fpm config issue before send you a PR. Any suggest about that?

Thanks!

geerlingguy:
Unfortunately no - don't have time to dig into it right now :(

But keep updating this issue with specific questions/ideas, or submit a PR and we'll continue from there. Thanks!

maschmann:
I think this has to do also with the php.ini placement. This changed from php version >=5.5 (could also have been 5.4) and now splits into two directories:
/etc/php5/cli and /etc/php5/fpm 
Thos directories have their own php.ini and respectively their own conf.d dir. In previous versions there was a symlink for conf.d directing to /etc/php5/conf.d
btw. this is causing the role to not overwrite the default config (php.ini) for php-fpm. Just stumbled upon that.

geerlingguy:
The general config for CLI/FPM/web is fixed now, as a result of #83—however, the root issue here (setting the value to `0` doesn't turn off opcache) is still not resolved, so leaving this ticket open.

geerlingguy:
This has been fixed via #131.

PR #86, opened 2016-02-29
Issue #48: Fix opcache/apc always enabled regardless of config

oxyc:
Havent actually tested but according to the `bool` implementation it should solve it. https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/filter/core.py#L115

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,20 +2,6 @@
 sudo: required
 
 env:
-  # Only test source install on latest supported OSes.
-  - distribution: centos
-    version: 7
-    init: /usr/lib/systemd/systemd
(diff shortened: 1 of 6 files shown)


oxyc:
Okay so I was on the wrong track entirely. This needs more thought, which is why it hasnt been tackled before I guess.

oxyc:
Works for Ubuntu but not Centos. This might not be the correct approach anyways...

oxyc:
Works for Ubuntu but not Centos. This might not be the correct approach anyways...

geerlingguy:
Should be fixed via #131.

PR #89, opened 2016-03-04
Apply configurations to all possible conf directories on Debian

oxyc:
https://github.com/geerlingguy/ansible-role-php-xdebug/pull/26#issuecomment-192050467


tasks/configure.yml
```diff
@@ -4,38 +4,44 @@
...
 
 - name: Place PHP configuration file in place.
   template:
     src: php.ini.j2
-    dest: "{{ php_conf_path }}/php.ini"
+    dest: "{{ item }}/php.ini"
     owner: root
     group: root
     mode: 0644
+  with_items:
```
geerlingguy: Since it's just one item, let's just use the shorthand:

```yaml
  with_items: "{{ php_conf_paths }}"
```

tasks/configure.yml
```diff
@@ -4,38 +4,44 @@
...
 - name: Place APC configuration file in place.
   template:
     src: apc.ini.j2
-    dest: "{{ php_extension_conf_path }}/{{ php_apc_conf_filename }}"
+    dest: "{{ item }}/{{ php_apc_conf_filename }}"
     owner: root
     group: root
     force: yes
     mode: 0644
+  with_items:
```
geerlingguy: Since it's just one item, let's just use the shorthand:

```yaml
  with_items: "{{ php_extension_conf_paths }}"
```

tasks/main.yml
```diff
@@ -13,15 +13,15 @@
     php_webserver_daemon: "{{ __php_webserver_daemon }}"
   when: php_webserver_daemon is not defined
 
-- name: Define php_conf_path.
+- name: Define php_conf_paths.
   set_fact:
-    php_conf_path: "{{ __php_conf_path }}"
-  when: php_conf_path is not defined
+    php_conf_paths: "{{ __php_conf_paths }}"
+  when: php_conf_paths is not defined
```
geerlingguy: This will probably need to be wrapped in {{}} so no DEPRECATION warning pops up.

tasks/main.yml
```diff
@@ -13,15 +13,15 @@
...
 
-- name: Define php_extension_conf_path.
+- name: Define php_extension_conf_paths.
   set_fact:
-    php_extension_conf_path: "{{ __php_extension_conf_path }}"
-  when: php_extension_conf_path is not defined
+    php_extension_conf_paths: "{{ __php_extension_conf_paths }}"
+  when: php_extension_conf_paths is not defined
```
geerlingguy: Same here with the {{}}

tasks/setup-Debian.yml
```diff
@@ -12,14 +12,18 @@
 
 - name: Delete APC configuration file if this role will provide one.
   file:
-    path: "{{ php_extension_conf_path }}/{{ php_apc_conf_filename }}"
+    path: "{{ item }}/{{ php_apc_conf_filename }}"
     state: absent
+  with_items:
```
geerlingguy: Same note as above with shorthand.

tasks/setup-Debian.yml
```diff
@@ -12,14 +12,18 @@
...
+    - "{{ php_extension_conf_paths }}"
   when: php_enable_apc and php_package_install.changed
   notify: restart webserver
 
 - name: Delete OpCache configuration file if this role will provide one.
   file:
-    path: "{{ php_extension_conf_path }}/{{ php_opcache_conf_filename }}"
+    path: "{{ item }}/{{ php_opcache_conf_filename }}"
     state: absent
+  with_items:
```
geerlingguy: Same note as above with shorthand.

vars/RedHat.yml
```diff
@@ -17,8 +17,11 @@ __php_packages:
   - php-xmlrpc
 __php_webserver_daemon: "httpd"
 
-__php_conf_path: /etc
-php_extension_conf_path: /etc/php.d
+__php_conf_paths:
```
geerlingguy: RHEL always so much simpler by default :)

tasks/main.yml
oxyc: Actually the warning is only for `with_*` and not conditionals.

vars/Debian.yml
```diff
@@ -14,10 +14,18 @@ __php_packages:
...
+__php_extension_conf_paths:
+  - /etc/php5/fpm/conf.d
+  - /etc/php5/apache2/conf.d
+  - /etc/php5/cli/conf.d
 
 __php_apc_conf_filename: 20-apcu.ini
 __php_opcache_conf_filename: 05-opcache.ini
 __php_fpm_daemon: php5-fpm
-__php_fpm_pool_conf_path: "/etc/php5/fpm/pool.d/www.conf"
+__php_fpm_conf_path: "/etc/php5/fpm"
```
oxyc: I had to add this variable because of Debain madness.

oxyc:
Hmmm:

Currently RedHat has:
```yml
__php_conf_path: /etc
__php_fpm_pool_conf_path: "/etc/php-fpm.d/www.conf"
```
But then this is in `php-fpm.conf`:
```
include={{ php_conf_path }}/fpm/pool.d/*.conf
```

oxyc:
Tests confirm that `--with-config-file-path` is mute. It's overridden anyways. The RedHat `pool.d` I still need to verify

tasks/main.yml
geerlingguy: Whoops! Misread on my part.

geerlingguy:
Looks good to me?

oxyc:
I'm not sure about the source install on RedHat. I'll run some tests to make sure

geerlingguy:
@oxyc - K, pending those results, I'll merge it in tonight or tomorrow morning. Working on a few other roles tonight.

oxyc:
Not sure I'll get it done tonight. Found some other bugs as well

oxyc:
Seems that the `/etc/php5/fpm/pool.d/www.conf` does not exist when installing from source with `--with-fpm`. Not on RedHat either. Did this use to work? I have to look into it more later

geerlingguy:
Don't have time to dig in now, but did we create that file using `template`?

oxyc:
We do not. It's available after installed from a package manager but not when installed from source.

geerlingguy:
I'm also looking into this now (a little at least) as I'm working on some [drupal-pi](https://github.com/geerlingguy/drupal-pi) improvements...

oxyc:
I had to give up. Think I need to install php 7.0 and play around properly. I think this was broken before this patch though. Currently I need a break from these crazy configs (I also read the paths have changed AGAIN for php7 on ubuntu when installed from the ppa...)

Here's a few notes from things I  attempted (still couldn't solve it though):

https://github.com/oxyc/ansible-role-php/commit/32d91a58c36e256d10654c42abb56a0dd3bcd4e3
Running [`systemctl status php-fpm.service` reported](https://travis-ci.org/geerlingguy/ansible-role-php/jobs/113733119#L1031) this path was the one included. Still same error though so maybe I'm wrong.

https://github.com/oxyc/ansible-role-php/commit/f248498f1346a1017eb3538ba4ddcfd1bccf4c25 Made sure the pool conf existed before modifying. Guess this is incorrect and we should add the file somehow.

https://github.com/oxyc/ansible-role-php/commit/5e116f15c552d88a39cad04f83dcd8aab463baae
Fixed the path to the conf files in `php-fpm.conf`. RedHat had it incorrect.

https://github.com/oxyc/ansible-role-php/commit/72873e890e30cf258a114757eb64fe9e105f3c89
I think this was broken because the variable was used in `install-from-source.yml` already.

geerlingguy:
It was definitely broken before this PR; I'm using that for testing on Debian Jessie, and it's breaking at the same spot. So I'll see if I can at least get that fixed.

I think we're safe to merge this, then fix the install-from-source issues in a separate PR, what do you think?

oxyc:
I think it's safe. There's one questionable commit that might affect php7 with fpm though 9eadfc1cdcb4f9d8a354133266dbbd7e312d391b, this change was what prompted me to test php7.

PR #91, opened 2016-03-04
Fixes variable name to make it configurable in RedHat

pedronofuentes:
`php_extension_conf_path` overrides the one you configure in your vars.

Changed to `__php_extension_conf_path` as it is in Debian.

### DIFF
--- a/vars/RedHat.yml
+++ b/vars/RedHat.yml
@@ -18,7 +18,7 @@ __php_packages:
 __php_webserver_daemon: "httpd"
 
 __php_conf_path: /etc
-php_extension_conf_path: /etc/php.d
+__php_extension_conf_path: /etc/php.d
 
 __php_apc_conf_filename: 50-apc.ini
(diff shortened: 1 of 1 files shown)


geerlingguy:
Thanks for the fix! This will be fixed in a separate PR from @oxyc - see https://github.com/geerlingguy/ansible-role-php/pull/89/files#diff-3f3da5a0b9c0c56c7f087e30468507d9R23 — I'm planning on merging that soon, so I'll close this PR and direct efforts towards that one (it fixes a few other path-based issues as well).

pedronofuentes:
Ok, no problem. I'll use my fork until the PR is merged.

Thank you for your work :)

geerlingguy:
@pedronofuentes - This should be fixed now, and it's a list (instead of one value) with just one value by default: https://github.com/geerlingguy/ansible-role-php/blob/master/vars/RedHat.yml#L23

Issue #93, opened 2016-03-08
phpenmod/phpdismod on Ubuntu


thom8:
Currently this role creates static `.ini`'s for php modules for apache, fpm & cli, so modules can't be easily enabled and disabled with this helper script.

Would need to create the static common `.ini` files @ `/etc/php5/mods-available` and then symlink them to the `/conf.d` folders, `phpenmod` could be used to do the symlinking, depending on OS support.

example >>

```
php5dismod xdebug
WARNING: Module xdebug ini file doesn't exist under /etc/php5/mods-available
WARNING: Module xdebug ini file doesn't exist under /etc/php5/mods-available
WARNING: Module xdebug ini file doesn't exist under /etc/php5/mods-available
``` 

relates to https://github.com/geerlingguy/ansible-role-php-xdebug/issues/22

geerlingguy:
Here's the thing... I detest a2enmod, php5dismod, et all—the whole idea of using strange platform/OS-specific shell script wrapper tools (which are version specific as well, meaning increased maintenance burden) irks me, because I'd rather manage all configuration via idempotent tools like Ansible.

On RedHat, Debian, etc. to disable an extension, you can just remove the config file, or comment out the `extension=` line, then restart the service.

I don't want to modify (and make even _more_ complex) the role config path layout just to make it better for some use cases on Debian/Ubuntu...

Sorry for the rant, but I just want to be clear that I am happy to do something to make enabling/disabling xdebug easier, but not at the expense of making things more complex, and definitely not in a way that's Debian/Ubuntu-specific.

thom8:
Makes complete sense but at the same time not following the standards in place for the platform/OS could potentially lead to other issues + this actually might make it less complex for supporting multiple versions by having a single module `ini` file symlinked to the multiple locations.

thom8:
Further to this and I'm not sure how it works on the RHEL, CentOS side of things but a common way to manage customisations to the `php.ini` file is with a custom ini in the `conf.d` dir.

This is often called `z_custom.ini` or similar so it's loaded last and can override defaults set in both the module and `php.ini`. Then you never need to modify the default files and wouldn't need to provide a full `php.ini` as a template.

There maybe be some good reasons not to do this but it could also help with managing multiple versions.

Happy to put some PR's together if it helps explain but wanted to discuss first.

thom8:
[...]

PR #96, opened 2016-03-20
Enable php modules with php5enmod on Debian.

thom8:
As mentioned in #93 this PR follows best practices when enabling PHP modules on Debian.

Changes to module config can be centrally managed from the `mods-available` directory rather than duplicating the `ini` files to each sapi conf directory.

```
Using these maintainer scripts will help get rid of the condition
when the package is removed, but not purged, hence triggering error
messages from php from the fact that it cannot load deleted dynamic
extension.
```

### DIFF
--- a/vars/Debian.yml
+++ b/vars/Debian.yml
@@ -20,12 +20,10 @@ __php_conf_paths:
   - /etc/php5/cli
 
 __php_extension_conf_paths:
-  - /etc/php5/fpm/conf.d
-  - /etc/php5/apache2/conf.d
-  - /etc/php5/cli/conf.d
+  - /etc/php5/mods-available
 
-__php_apc_conf_filename: 20-apcu.ini
-__php_opcache_conf_filename: 05-opcache.ini
+__php_apc_conf_filename: apcu.ini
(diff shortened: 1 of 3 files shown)


vars/Debian.yml
```diff
@@ -20,12 +20,10 @@ __php_conf_paths:
   - /etc/php5/cli
 
 __php_extension_conf_paths:
-  - /etc/php5/fpm/conf.d
-  - /etc/php5/apache2/conf.d
-  - /etc/php5/cli/conf.d
+  - /etc/php5/mods-available
```
thom8: Whilst this PR implements `php5enmod` which is Debian/Ubuntu specific, you don't need to manage all possible Debian/Ubuntu specific `php_extension_conf_paths`. 

All installed PHP sapi's will be automatically linked, with the correct priority when the module is enabled via `php5enmod`.

thom8:
Closing as this method is not supported by Ubuntu 12.04.

geerlingguy:
See also: https://github.com/geerlingguy/ansible-role-php/issues/93#issuecomment-199501176

PR #97, opened 2016-03-20
Use zzz_custom.ini to override default php.ini.

thom8:
Follow on from #93 

The default `php.ini` doesn't ever never need to be modified as the dynamic conf directories can be used to manage overrides.

This PR replaces the current `php.ini` template with a `zzz_custom.ini` file which is placed in the dynamic conf directory, the `zzz` prefix is to ensure this is loaded last and will override all other config.


thom8:
Closing out old PRs.

PR #100, opened 2016-03-26
Add basic FreeBSD support

cmprescott:
This isn't as thorough as your Debian & EL support, but I figure others might want basic PHP on FreeBSD too. Currently I couldn't spy any even with sorting through less popular roles in the Ansible Galaxy.


geerlingguy:
Closing old/failed build PRs; please open a new one if you can get a chance. Note that I don't use or test on *BSD, so I wouldn't be able to offer any extra support for it.

PR #103, opened 2016-04-08
Typo in php.ini mssql.compatibility_mode

Petrox:
Bad:
mssql.compatability_mode = Off

Good:
mssql.compatibility_mode = Off

### DIFF
--- a/templates/php.ini.j2
+++ b/templates/php.ini.j2
@@ -209,7 +209,7 @@ mssql.max_persistent = -1
 mssql.max_links = -1
 mssql.min_error_severity = 10
 mssql.min_message_severity = 10
-mssql.compatability_mode = Off
+mssql.compatibility_mode = Off
 mssql.secure_connection = Off
 
 [Tidy]


geerlingguy:
Heh, from: https://bugs.php.net/bug.php?id=53941 (see also: https://bugs.php.net/bug.php?id=69052).

But I like spelling things correctly, so I'll merge this in :).

Petrox:
But if the php config option name has the typo we have to stay that way,
and my pull request is invalid then...

Also is there :facepalm here?

geerlingguy:
@Petrox - It works either way (the typo is basically an alias to the correct spelling). So either way works as well as the other, at least as far back as PHP 5.2 (probably earlier).

geerlingguy:
Hmm... digging deeper it looks like 5.4 still has `compatability` only, so I'm going to revert the commit... since some people still use it :P

Petrox:
Thank you, and sorry. :)

geerlingguy:
No problem! If it weren't for PHP 5.4, I'd rather have it spelled correctly :P

PR #112, opened 2016-04-14
Restart php-fpm when apc/opcodecache configs change

andreiashu:
Just making sure that php-fpm is restarted when the apc/opcodecache configs change.

### DIFF
--- a/tasks/configure.yml
+++ b/tasks/configure.yml
@@ -29,7 +29,9 @@
     mode: 0644
   with_items: "{{ php_extension_conf_paths }}"
   when: php_enable_apc
-  notify: restart webserver
+  notify:
+    - restart webserver
+    - restart php-fpm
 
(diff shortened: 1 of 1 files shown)


geerlingguy:
Hmm... I'm torn on this—the current config (php.ini) task doesn't do both, but instead calls `restart webserver` and then that, in turn calls `restart php-fpm`... but it's probably best to have both handlers called in `notify` for all _three_ of the config files, because if you're not using a webserver that needs a restart (e.g. `php_enable_webserver` is `false`), then that handler won't run, and it won't notify `restart php-fpm` :/

So I guess, could you add both handlers to all three config file changes (now in `configure.yml`, `configure-apcu.yml`, and `configure-opcache.yml`)?

geerlingguy:
Can you please rebase and create a new PR? I'm okay with merging as long as we cover all the bases here. Thanks!

PR #113, opened 2016-04-14
Enable opcache.revalidate_path setting

andreiashu:
Add ability to enable opcache.revalidate_path setting


defaults/main.yml
```diff
@@ -26,6 +26,7 @@ php_opcache_interned_strings_buffer: "16"
 php_opcache_max_accelerated_files: "4096"
 php_opcache_max_wasted_percentage: "5"
 php_opcache_validate_timestamps: "1"
+php_opcache_revalidate_path: "1"
```
geerlingguy: I'd rather keep the default the same as the out-of-the-box default (`0`); see http://files.zend.com/help/Zend-Server/content/zendserverapi/zend_opcache-configuration_directives.htm#dir-opcache.revalidate_path

Then users can override if they see fit.

PR #118, opened 2016-04-27
Add: Ubuntu support for install_from_source task

gavinrogers:
There is one dependancy with a different name in Ubuntu (libreadline-dev) as well as needing an upstart script and a few other little bits and pieces.

### DIFF
--- a/tasks/install-from-source.yml
+++ b/tasks/install-from-source.yml
@@ -78,7 +78,7 @@
 
 - name: Clone the PHP repository.
   git:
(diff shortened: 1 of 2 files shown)


geerlingguy:
Hmm... I currently have the install from source tests run via Travis on both Ubuntu 12 and 14, and I've also tested install from source on Ubuntu 14 and 16, as well as Debian 7 and 8, and everything seems to work with the defaults currently provided.

PR #128, opened 2016-05-19
Fix #120: Introduce variables for FPM user and group.

### LINKED ISSUE #120: PHP-FPM fails on Centos because of wrong configuration.
PHP-FPM config file www.conf sets user and group for apache to "www-data" but in Centos Apache user and group are "apache" instead. This causes PHP-FPM start to fail, because it tries to get UUID of user that does not exist.

mgarstecki:
Fix issue #120, by introducing variables for the FPM pool user and group, and templating the default pool configuration to pull them in.

Comments and suggestions welcome.

### DIFF
--- a/vars/RedHat.yml
+++ b/vars/RedHat.yml
@@ -28,3 +28,6 @@ __php_opcache_conf_filename: 10-opcache.ini
 __php_fpm_daemon: php-fpm
 __php_fpm_conf_path: "/etc/fpm"
 __php_fpm_pool_conf_path: "/etc/php-fpm.d/www.conf"
+
+__php_fpm_pool_user: apache
+__php_fpm_pool_group: apache
--- a/tasks/configure-fpm.yml
+++ b/tasks/configure-fpm.yml
@@ -9,6 +9,16 @@
     php_fpm_pool_conf_path: "{{ __php_fpm_pool_conf_path }}"
   when: php_fpm_pool_conf_path is not defined
 
+- name: Define php_fpm_pool_user.
+  set_fact:
+    php_fpm_pool_user: "{{ __php_fpm_pool_user }}"
+  when: php_fpm_pool_user is not defined
+
+- name: Define php_fpm_pool_group.
+  set_fact:
+    php_fpm_pool_group: "{{ __php_fpm_pool_group }}"
+  when: php_fpm_pool_group is not defined
+
 - name: Stat php_fpm_pool_conf_path
   stat:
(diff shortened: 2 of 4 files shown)


vars/RedHat.yml
```diff
@@ -28,3 +28,6 @@ __php_opcache_conf_filename: 10-opcache.ini
 __php_fpm_daemon: php-fpm
 __php_fpm_conf_path: "/etc/fpm"
 __php_fpm_pool_conf_path: "/etc/php-fpm.d/www.conf"
+
+__php_fpm_pool_user: apache
```
geerlingguy: Hmm... it would only be `apache` if you're using Apache as a webserver, right?

I have `www-data` set on my Drupal VM instance running CentOS 7, and it seems to work okay.

valdars: Yes. Also i think there was another condition for this problem to occur. Can't remember and i am not currently working with this project. Maybe enable FPM?

mgarstecki: @valdars : indeed, for me the problem only occurs when `php_enable_php_fpm` is `true`. That starts the php-fpm service, which fails when the pool user/groups are not defined. 

@geerlingguy :
Strange, I don't have the `www-data` user on my fresh CentOS 7 instances. What apache package do you use ?

Since the php RPM depends on httpd, the `apache` user will be available on most php installs, and seemed a good default to me. I intend to pair PHP with nginx, and it seems this should work even if they don't use the same users.

What default value would you suggest ?

mamchenkov: The other condition might be the ownership of the session directory ... something like `/var/lib/php/5.6/` ... it has to be server writeable or any application using sessions (which is pretty much every application these days) will fail.  File storage for sessions is the default in PHP.

mamchenkov: Can we also have the variable for the group.  www.conf has both user and group values, so it would be nice to be able to control both and not just the user half. :smile:

mgarstecki: @mamchenkov :
* Session directory : I'm currently testing a PHP app with this configuration, we'll see how it goes.
* Variable for the group : already handled in the PR :)

geerlingguy:
So, all tests pass fine, so I'm going to test things out a little further locally and make sure it's all good, then this looks safe to merge.

vars/RedHat.yml
geerlingguy: You seem to be correct on the `apache` vs `www-data` thing. Don't know where I was getting `www-data` from!

PR #131, opened 2016-05-20
Fixes #130 and #48: Tweak and simplify some of the apc and opcache configuration

### LINKED ISSUE #130: CLI apc/apcu extension warnings.
Following up after #125.

Using version 3.0.4 and DrupalVM with geerlingguy/1404.

If php_apc_enabled_in_ini: false we see this warning since apcu extension is not loaded.

```PHP Warning:  PHP Startup: Unable to load dynamic library '/usr/lib/php/20151012/apc.so' - /usr/lib/php/20151012/apc.so: undefined symbol: zif_apcu_store in Unknown on line 0```

If php_apc_enabled_in_ini: true we see this warning since apc extension is also loaded in 20-apcu_bc.ini.
```PHP Warning:  Module 'apc' already loaded in Unknown on line 0```

It appears php-apcu-bc is automatically installed.  Not sure the best solution for a PR.

rockholla:
[Related to this issue](https://github.com/geerlingguy/ansible-role-php/issues/130) and [this one](https://github.com/geerlingguy/ansible-role-php/issues/48), I thought some tweaks needed to be made with configuration file management for apc and opcache.  The fix is simply to remove any apc or opcache configuration files that the PHP installation creates in favor of managing the configuration files through this role.

Another change is to remove the variables `php_opcache_enabled_in_ini` and `php_apc_enabled_in_ini` and just rely on the switches `php_opcache_enable` and `php_enable_apc` to create or remove the relevant config files.  I think this should suffice unless I'm missing something.  Just let me know.

For the sake of noting everything related going on the repo, it looks like [this pull request](https://github.com/geerlingguy/ansible-role-php/pull/86) was a attempting to solve a similar issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,10 +87,6 @@ Various defaults for PHP. Only used if `php_use_managed_ini` is set to `true`.
 
 The OpCache is included in PHP starting in version 5.5, and the following variables will only take effect if the version of PHP you have installed is 5.5 or greater.
 
-    php_opcache_enabled_in_ini: false
-
-When installing Opcache, depending on the system and whether running PHP as a webserver module or standalone via `php-fpm`, you might need the line `extension=opcache.so` in `opcache.ini`. If you need that line added (e.g. you're running `php-fpm`), set this variable to true.
-
     php_opcache_enable: "1"
     php_opcache_enable_cli: "0"
     php_opcache_memory_consumption: "96"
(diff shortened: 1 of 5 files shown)


geerlingguy:
Yeah, this is slightly more involved, of course, than the approach #48 was taking, but I think it will solve a lot of the small/painful issues people have been running into with different PHP versions on different platforms.

I'm going to merge this, then maybe do a minor release and hopefully people can test and see how it holds up (I don't think it should break anything—but in some cases this role's configuration might not have been used before this PR).

rockholla:
Excellent, thanks @geerlingguy

PR #132, opened 2016-05-24
Add tests for php warnings, apcu and opcache extensions.

joestewart:
Additional tests for php warnings, and whether apcu and opcache extensions were loaded.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -77,6 +77,12 @@ script:
   # Ensure PHP configurations have taken effect.
   - sudo docker exec --tty "$(cat ${container_id})" env TERM=xterm php -i | grep 'memory_limit.*192'
 
+  # Ensure php apcu extension loaded
+  - sudo docker exec --tty "$(cat ${container_id})" env TERM=xterm php -i | grep -i 'apc.enabled'
+
+  # Ensure php opcache extension loaded
+  - sudo docker exec --tty "$(cat ${container_id})" env TERM=xterm php -i | grep 'opcache.enable'
+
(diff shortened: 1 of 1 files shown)


joestewart:
The tests do seem to show the differing failures for OS and php versions.

Will be nice to get Ubuntu 1604 in the matrix too.  Is the current work on this in the mysql role?

geerlingguy:
I'm wondering - do you think it would be good to have those tests in permanently? They may be a little brittle in terms of some of the older tests... but apcu and opcache should be present/working in any modern version of PHP installed by this role.

joestewart:
I thought so, the tests run quickly and with the default role configuration apcu and opcache should be loaded and have no warnings.

The reason I submitted this is that we saw different failures with OS and php version combinations

joestewart:
Ok, I rebased after #136 fixed up part of the apc/apcu issues.  This helps bring some clarity to the failures.

The failures seem to be:

1. php source installation doesn't enable apcu nor opcache.
2. ubuntu 12.04 doesn't enable apcu nor opcache.

All of these may be acceptable here even though the default variables imply they would be enabled.
So this PR likely can be closed.

geerlingguy:
Yeah; for source install, it's up to the user to provide extra flags to add support for additional capabilities, though we should include opcache support OOTB. And for Ubuntu 12 and CentOS 6, you can update packages using extra repos to PHP 5.5+, and the defaults should work better. I'm not going to be supporting PHP < 5.5 with this role anymore as of 3.0.x.

PR #138, opened 2016-06-22
Add explicit configuration for php-fpm

jeremylivingston:
If the current `php_enable_php_fpm` variable is set to true, php-fpm
will be both configured and started. This may not be the desired outcome
for those who want to configure php-fpm without starting the process.

Add a new variable, `php_start_php_fpm`, that can be configured to
handle the process behavior. This will set to the value of
`php_enable_php_fpm` by default to maintain backwards compatibility,
but can be modified as desired.

### DIFF
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -10,4 +10,4 @@
   service:
     name: "{{ php_fpm_daemon }}"
     state: restarted
-  when: php_enable_php_fpm
+  when: php_start_php_fpm
--- a/tasks/configure-fpm.yml
+++ b/tasks/configure-fpm.yml
@@ -73,4 +73,4 @@
     name: "{{ php_fpm_daemon }}"
     state: started
     enabled: yes
-  when: php_enable_php_fpm
+  when: php_start_php_fpm
--- a/README.md
+++ b/README.md
@@ -46,6 +46,11 @@ When using this role with PHP running as `php-fpm` instead of as a process insid
 
     php_enable_php_fpm: false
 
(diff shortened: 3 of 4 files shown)


geerlingguy:
Just to make sure I'm clear on the idea—what would be a case where you'd use this role and install php-fpm but not want it to be running at the end?

jeremylivingston:
@geerlingguy My particular use case is that I have a deployment process that kicks off when a new server is started up. I don't want php-fpm to be running until my deployment is completed because it would be added to my pool prematurely. The deployment process handles the responsibility of starting and stopping the process at different points in the lifecycle.

jeremylivingston:
Any thoughts on this, @geerlingguy?

handlers/main.yml
```diff
@@ -10,4 +10,4 @@
   service:
     name: "{{ php_fpm_daemon }}"
     state: restarted
-  when: php_enable_php_fpm
+  when: php_start_php_fpm
```
TroodoNmike: this should be:
`when php_enable_php_fpm and php_start_php_fpm`

tasks/configure-fpm.yml
```diff
@@ -73,4 +73,4 @@
     name: "{{ php_fpm_daemon }}"
     state: started
     enabled: yes
-  when: php_enable_php_fpm
+  when: php_start_php_fpm
```
TroodoNmike: same as above

TroodoNmike:
I agree this change is useful. I have post tasks that do something before fpm is started

nonotest:
+1

PR #139, opened 2016-07-07
Update Debian.yml

mheesters:
php-apcu package doesn't exist, the php7.0-apcu does

### DIFF
--- a/vars/Debian.yml
+++ b/vars/Debian.yml
@@ -13,7 +13,7 @@ __php_packages:
   - php7.0-xml
   - php7.0-mbstring
   - php-sqlite3
-  - php-apcu
+  - php7.0-apcu
 __php_webserver_daemon: "apache2"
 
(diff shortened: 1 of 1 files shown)


geerlingguy:
@mheesters - In the common repo for Xenial, it seems only `php-apcu` is available, but not `php7.0-apcu`: http://packages.ubuntu.com/xenial/php/php-apcu

Since this role's lowest common denominator is the system packages, I base the package list on what's available there... IIRC, the `php7.0-apcu` package is only available if using a PPA like Ondrej's, right?

mheesters:
I see. I was installing it on Debian (8) so for php7 apcu i had to use the dotdeb archive.
https://www.dotdeb.org/instructions/

geerlingguy:
@mheesters - In that case, it's probably a good idea for you to keep the `php_packages` variable in your playbook's vars (to override this default).

PR #142, opened 2016-07-10
Make sure we can clone PHP repository using git

TroodoNmike:
Straight forward change. I am using php build from source. When I install playbook for plain/clean ubuntu it fails as git cannot clone repo. I have added task to make sure we can clone php repo using git.

### DIFF
--- a/tasks/install-from-source.yml
+++ b/tasks/install-from-source.yml
@@ -76,6 +76,10 @@
   failed_when: false
   register: php_installed
 
+- name: Make sure we can clone PHP repository using git.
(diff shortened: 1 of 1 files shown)


geerlingguy:
@TroodoNmike - I'd much rather the installation of Git itself be managed by the playbook that uses this role—for a simple multi-OS install (with options for building Git from source), check out [`geerlingguy.git`](https://galaxy.ansible.com/geerlingguy/git/). You can just add that role before `geerlingguy.php` and everything should work well.

PR #144, opened 2016-07-26
Add bcmath module for php 7

valentine94:
Hi Jeff.

I think we need this PHP module to be installed, because it's required for some modules(in my case - I'm just can't install commerce on D8).

(I'm using DrupalVM with Ubuntu 16.04 box)

What do you think?

Thanks.

### DIFF
--- a/vars/Debian.yml
+++ b/vars/Debian.yml
@@ -1,5 +1,6 @@
 ---
 __php_packages:
+  - php7.0-bcmath
   - php7.0-common
   - php7.0-cli
   - php7.0-dev


geerlingguy:
@valentine94 - In this particular case, since this module is only required by a (comparatively) small number of Drupal sites, I would recommend just setting the list of `php_packages` in your Drupal VM's `config.yml` file.

This PHP role is used by thousands (maybe hundreds of thousands?) of PHP projects/sites (not just Drupal), so I try to keep the default set of packages as generic as possible to support a broad range of use cases but not require installing a ton of unnecessary extensions.

valentine94:
OK, that makes sense, thanks.

PR #145, opened 2016-07-26
Enables repos for Debian apt distro

dustinalan:
Worked for Centos/Yum but not on Debian/Ubuntu/Apt

### DIFF
--- a/tasks/setup-Debian.yml
+++ b/tasks/setup-Debian.yml
@@ -1,4 +1,7 @@
 ---
+- name: Enable Repo
+  apt_repository: repo={{ php_enablerepo }}
+  
 - name: Update apt cache.
   apt: update_cache=yes cache_valid_time=86400
 


geerlingguy:
If you need a more current apt PPA for Ubuntu (or a different repo for other distros), then the best way to use that is to add a separate task in `pre_tasks` that runs prior to this role, for example, here's how Drupal VM does it:

https://github.com/geerlingguy/drupal-vm/blob/master/provisioning/tasks/init-debian.yml#L23-L33

dustinalan:
Nice, so many ways to do the same thing :+1:

dustinalan:
However, the variable `php_enablerepo` is unused with Ubuntu, which I think needs a comment here:

https://github.com/geerlingguy/ansible-role-php/blob/master/defaults/main.yml#L3

I like the drupal options between the 3 releases, thanks for that.

geerlingguy:
@dustinalan - Good idea! It was documented in the main README file, but it would help to also document it in defaults, so I've done it there.

dustinalan:
Keep up the good work! Learned a lot reading through your playbooks :-) :+1:

PR #147, opened 2016-08-22
Ensure php-fpm is stopped when defined

thiagomgo:
Ensure that php-fpm is stopped and disabled, once that it already installed on the server.

### DIFF
--- a/tasks/configure-fpm.yml
+++ b/tasks/configure-fpm.yml
@@ -74,3 +74,10 @@
     state: started
     enabled: yes
   when: php_enable_php_fpm
+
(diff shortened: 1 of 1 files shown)


geerlingguy:
@thiagomgo - For your own playbooks, if you're transitioning from FPM to something else, you can add a task to do the opposite of what setting `php_enable_php_fpm` does (https://github.com/geerlingguy/ansible-role-php/blob/master/tasks/configure-fpm.yml#L71-L76), or you can rebuild the server without FPM enabled. That's generally the model I follow for my roles, so I don't have to have two sets of tasks for any option (e.g. one to set something up, another to turn something off).

PR #153, opened 2016-10-05
Added php option for disable_functions

stevenspasbo:
- Added php_disable_functions variable
- Updated php.ini template


stevenspasbo:
I was getting annoyed by this [issue](https://github.com/drush-ops/drush/issues/2065). This change allowed me to get drush to stop complaining.

PR #154, opened 2016-10-14
Add php7 zip extension

stevenjlho:
After i upgrade ansible-role-php latest version, my php version is upgrade to 7.0 in Ubuntu 14.04. But I'm trying to use Composer to install Laravel on Ubuntu 14.04, i get something warning below:

```
Failed to download symfony/finder from dist: The zip extension and unzip command are both missing, skipping. The php.ini used by your command-line PHP is: /etc/php/7.0/cli/php.ini
```

I find that should install zip extension. Laracasts  have a similar [issue](https://laracasts.com/discuss/channels/laravel/zip-php-extension-problem-in-ubuntu-php7?page=1). So i want to add php zip extension.

### DIFF
--- a/vars/Debian.yml
+++ b/vars/Debian.yml
@@ -12,6 +12,7 @@ __php_packages:
   - php7.0-opcache
   - php7.0-xml
   - php7.0-mbstring
+  - php7.0-zip
   - php-sqlite3
   - php-apcu
 __php_webserver_daemon: "apache2"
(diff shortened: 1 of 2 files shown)


geerlingguy:
@stevenjlho - Thanks so much for the PR! I'd rather not add the `zip` extension to the default packages list, since most projects I work on don't _explicitly_ require it, and even with Composer, you can set things up to not require it either. But as a rule, most larger projects will likely need to (and it would be best to) lock in their own `php_packages` set anyways, so that's what I'd recommend here.

PR #156, opened 2016-10-25
Set php_expose_php to be "Off" by default

Brunty:
Being on by default seems like a bad idea as people will forget to set it off themselves which should be the default (IMO)

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -47,7 +47,7 @@ php_apc_enable_cli: "0"
 # Any and all changes to /etc/php.ini will be your responsibility.
 php_use_managed_ini: true
 
-php_expose_php: "On"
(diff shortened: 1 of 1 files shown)


geerlingguy:
We could have a nice argument about security through obscurity and whether it's more helpful to force the hiding of this information on users or not... but for 99% of use cases, I like to stick to PHP's (and other systems') defaults so users aren't surprised when they use the role the first time. And in this case, PHP still prints its info by default: http://php.net/manual/en/ini.core.php

So if it's changed upstream to "Off"/`0`, then I'll change it here. But there are many valid reasons for exposing the PHP info—one of my favorite is to shame people who are running older versions into upgrading (that's why we automate!) :)

PR #157, opened 2016-11-07
add support to set custom opcache.so location

jamonkko:
Here is the original problem for anyone having the same issue: There was this error when trying to run php in Centos7. 
`Failed loading opcache.so:  opcache.so: cannot open shared object file: No such file or directory`

Fixed it by installing php56-php-opcache from remi-php56 repo and specifying full path as zend_extension:
`php_opcache_zend_extension: "/opt/remi/php56/root/usr/lib64/php/modules/opcache.so"`


geerlingguy:
Build failure is related to https://github.com/geerlingguy/ansible-role-php/issues/158 - I'll try to get that fixed separately.

geerlingguy:
This is in the 3.4.0 release. Thanks!

PR #162, opened 2016-11-18
Enable service for php-fpm

andres-ortiz:


geerlingguy:
Ansible's `service` module should already be doing this correctly (as long as you're running a recent version of Ansible)—it should work either on older sysvinit, or on systemd.

Were you running into a particular error?

andres-ortiz:
My error for build php is:

```TASK [geerlingguy.php : Ensure php-fpm is started and enabled at boot (if configured).] ***
fatal: [default]: FAILED! => {"changed": false, "failed": true, "msg": "Could not find the requested service \"'php-fpm'\": "}``` 

but if i add systemctl php-fpm enable is good.

ansible 2.2.0.0

**Comment by geerlingguy** [comment]
@andres-ortiz - Can you see (just so I can confirm a suspicion) if you let the role fail, then restart your server, then run the role again—if it works the second time after a restart?

If so, then I think this is related to a weird bug that occurs with some older sysvinit scripts that don't work too well on systemd systems (usually just Ubuntu 16.04 for some reason—CentOS 7 and Debian 8 aren't always affected).

Also, if you can run Ansible 2.1.x and try there, and if the error isn't present, I'm pretty sure it's this bug.

**Comment by mglaman** [comment]
@geerlingguy having same issue.

```
TASK [geerlingguy.php : Ensure php-fpm is started and enabled at boot (if configured).] ***
fatal: [localhost]: FAILED! => {"changed": false, "failed": true, "msg": "Could not find the requested service \"'php7.0-fpm'\": "}
	to retry, use: --limit @/root/playbook.retry
```

```
~# ansible --version
ansible 2.2.0.0
  config file = /etc/ansible/ansible.cfg
  configured module search path = Default w/o override
```

Fresh DigitalOcean Ubuntu 16 droplet. Rebooted and nothing.

Have not tried rolling to 2.1. Can confirm it works, though, on a Ubuntu 14 droplet.

geerlingguy:
@mglaman - Thanks for the confirmation.

So far I'm unable to reproduce this error in any of my production environments (almost all running either CentOS 7 or Ubuntu 16.04), nor have I heard any reports of it failing for any of the Drupal VM or other downstream projects that default to 16.04. So closing for now, but if you can give more specific steps to reproduce, I'll be happy to look into it further.

PR #165, opened 2016-12-01
Adds variable additional_php_packages

solonair:
Allows to add additional php packages without overriding the default.


geerlingguy:
Added another commit—the variable name should be prefixed with `php_`, so I chose `php_packages_extra` and updated the README.

Thanks!

PR #168, opened 2017-01-10
Changed default value of short_open_tags to Off instead of False

yaron:
https://secure.php.net/manual/en/ini.core.php#ini.short-open-tag

The value should be "Off" instead of "False". This was changed in the pull request making the variable configurable: https://github.com/geerlingguy/ansible-role-php/pull/7/files

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -61,7 +61,7 @@ php_allow_url_fopen: "On"
 
 php_sendmail_path: "/usr/sbin/sendmail -t -i"
 php_output_buffering: "4096"
-php_short_open_tag: false
+php_short_open_tag: "Off"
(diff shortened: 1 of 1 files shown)


geerlingguy:
Technically, shouldn't it be `0` to match the docs? http://php.net/manual/en/ini.core.php

yaron:
I think they are all parsed to the same true/false in PHP internally, but thought I'd use the same format as used elsewhere in the php.ini.j2 file.

geerlingguy:
Hah, 'consistency'... 'php'... they don't often appear together :)

I'll merge as 'Off', since it makes more sense to me than 0 in terms of PHP settings flag consistency.

PR #172, opened 2017-01-25
skip default PHP-FPM pool creation when requested

VerosK:
In our use-case, we're creating php-fpm pools manually, so we don't need default php-fpm pool.

So I suggest to add `php_fpm_create_default_pool`  option that will disable creating of default pool.

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -21,6 +21,8 @@ php_fpm_pm_max_children: 50
 php_fpm_pm_start_servers: 5
 php_fpm_pm_min_spare_servers: 5
 php_fpm_pm_max_spare_servers: 5
+# Create default pool in php-fpm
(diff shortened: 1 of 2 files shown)


FinBoWa:
Isn't the default pool created by PHP installation this role just overrides that one. If you do this https://github.com/geerlingguy/ansible-role-php/pull/172/files#diff-092d14e309614f09e51c6c34253a9d2dR69 the file would be there but without any values from this role

VerosK:
@FinBoWa   I delete the default pool config in www.conf in some roles running after this one. But on next ansible run, this role created the default config again.  

So I suggest, it's better to add variable to not create the default pool config agian

VerosK:
Closing for now as not important for me.

Issue #98, opened 2016-03-20
php-fpm pool www.conf will be created even php-fpm isn't installed


StefanKlatt:
Hi,

php-fpm pool www.conf will be created even php-fpm isn't installed.

What about extending the the last task "# Configure PHP-FPM." with: 

 when: php_enable_php_fpm

?

Regards

Stefan

acie:
Hi,

I also got the same issue. Even on my defaults/main.yml the **php_enable_php_fpm: false**. PHP-FPM is still running.

Any advice?

Thanks.

geerlingguy:
True... the `when` would need to be added on [this task](https://github.com/geerlingguy/ansible-role-php/blob/master/tasks/configure-fpm.yml#L36), `Ensure the default pool exists.`. I may do this... no need to add the config if it's not being used.

geerlingguy:
This was fixed in the above commit. Thanks @FlxPeters !

PR #176, opened 2017-02-09
Fix issue #98 - Update configure-fpm.yml

FlxPeters:
Only add the default pool when php_enable_php_fpm is true


geerlingguy:
Looks good, thanks!

PR #179, opened 2017-03-01
add systemd support for php-fpm build from source

vvgelder:
Hi, I noticed that when I build php from source the php-fpm init script does not work on systemd. This is a possible fix. I used a separate restart php-fpm handler because systemd needs a daemon reload. Don't know if this is the best way to do this.


tasks/configure-fpm.yml
```diff
@@ -69,9 +69,16 @@
...
+    state: restarted
+  when: php_enable_php_fpm and ansible_service_mgr == 'systemd'
+
 - name: Ensure php-fpm is started and enabled at boot (if configured).
   service:
     name: "{{ php_fpm_daemon }}"
     state: started
     enabled: yes
-  when: php_enable_php_fpm
+  when: php_enable_php_fpm and ansible_service_mgr != 'systemd'
```
geerlingguy: Besides Debian, this should work correctly whether the service manager is systemd or initv.

tasks/configure-fpm.yml
```diff
@@ -69,9 +69,16 @@
   when: php_enable_php_fpm
   notify: restart php-fpm
 
+- name: systemd restart php-fpm
+  systemd:
+    name: "{{ php_fpm_daemon }}.service"
+    daemon_reload: yes
+    state: restarted
```
geerlingguy: Always restarting this service would cause the role's idempotence to break. It would need to only be done if triggered via a `notify` or something like that.

Also, the `service` module should automatically work with both initv and systemd, depending on what's available—so I still wonder if it would be necessary to do this. Maybe only if the task that copies the unit file reports `changed`?

PR #180, opened 2017-03-13
Change README to point to correct lineinfile hint

eiriksm:
Hey there. Just followed the tip about adding additional config to www.conf, but the instructions probably point to the wrong file? Anyway, I found it more helpful to look at configure-fpm.yml, so I am assuming this is what people should look at to alter that file.

Thanks for making this, btw!


geerlingguy:
Perfect, and yes, that was an old reference that needed updating. Thanks!

PR #193, opened 2017-05-16
PHP-FPM service not enabled on boot on Debian 8

oxyc:
Just checking what travis says. This particular fix would require Ansible 2.2 and be targeted at Debian only so it's a no go.


geerlingguy:
I cherry-picked your commit into my upstream branch, so closing this PR.

PR #196, opened 2017-05-19
Add Debian ppa support

wizhippo:


geerlingguy:
There's actually a role that's part of Drupal VM (though it's slated to be put into it's own role on Galaxy, see https://github.com/geerlingguy/drupal-vm/issues/1377) for PHP version switching: https://github.com/geerlingguy/drupal-vm/tree/master/provisioning/roles/drupalvm.php-versions

I'd rather focus on getting that role working well instead of adding more versioning support to this (already complex enough) role.

geerlingguy:
See: https://github.com/geerlingguy/ansible-role-php-versions

PR #200, opened 2017-05-29
Add functionality to load php5 modules on Debian systems

dthielking:
On Debian derivates mcrypt is not automatically loaded. So I wrote a task to enable loading of php modules. I hope that will fit in your role

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,10 @@ _Note: If you're using Debian/Ubuntu, you also need to install `libapache2-mod-f
 
 A list of extra PHP packages to install without overriding the default list.
 
+    php_load_modules: []
+
+A list of modules to load on Debian, if not automatically loaded.
+
     php_enable_webserver: true
 
 If your usage of PHP is tied to a web server (e.g. Apache or Nginx), leave this default value. If you are using PHP server-side or to run some small application, set this value to `false` so this role doesn't attempt to interact with a web server.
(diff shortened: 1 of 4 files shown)


dthielking:
@geerlingguy is it ok for you to merge it into upstream or do I have to do something?

geerlingguy:
I'm still thinking about this... since it's debian-specific, I'm not too sure I want to include it directly; it's similar to some of the apache2* utilities Debian/Ubuntu provide—if I can't support them across multiple OSes, I don't like the idea of making the role work one way on some systems, and another way on others.

dthielking:
Yeah, I see. But if this role should support more than one OS we have to deal with upstream tools from e.g. Debian/Ubuntu. I think If we add more support to this cases this role will be the best to use on all derivates.

dthielking:
@geerlingguy did you taken any decisions about this PR? I am now using the forked one, but i want to go with this upstream role. I hope you will understand.

dthielking:
No merged and won't be merged.

geerlingguy:
Sorry, I had a pretty crazy June, so I'm only starting to churn through the backlog of issues in my roles; I think I have to agree with my earlier statement that I'd rather not maintain this Debian/Ubuntu-specific feature, as it's (IMO) kind of unintuitive (why install modules you aren't using if all your infra operations are automated?); if you need the functionality, my best advice is to write a separate role that manages modules that way. If there's a small change to this role that could help make that role easier to maintain, I'd consider that.

PR #201, opened 2017-06-01
Set the correct opcache priority on Debian

ak4t0sh:
cf issue on geerlingguy.php-versions : https://github.com/geerlingguy/ansible-role-php-versions/issues/2

### DIFF
--- a/vars/Debian.yml
+++ b/vars/Debian.yml
@@ -28,7 +28,7 @@ __php_extension_conf_paths:
   - /etc/php/7.0/cli/conf.d
 
 __php_apc_conf_filename: 20-apcu.ini
-__php_opcache_conf_filename: 05-opcache.ini
+__php_opcache_conf_filename: 10-opcache.ini
 __php_fpm_daemon: php7.0-fpm
 __php_fpm_conf_path: "/etc/php/7.0/fpm"
 __php_fpm_pool_conf_path: "{{ __php_fpm_conf_path }}/pool.d/www.conf"


ak4t0sh:
@geerlingguy i don't know why php5.6 test fails on Ubuntu but do you want me to fix it or is it ok for you to merge ?

geerlingguy:
I think it actually defaults to 05 using the system packages... but I could be mis-remembering.

geerlingguy:
And I think that test is failing because nobody supports Ubuntu 12.04 anymore :) I should probably remove that test.

ak4t0sh:
As explained in https://github.com/geerlingguy/ansible-role-php-versions/issues/2

> This seems to occur because Ondrej's repo for Debian also sets the file as priority 10 and not 05.

So Ondrej's repo create 10-opcache.ini but this role create 05-opcache.ini this is why you remove the "when ubuntu" condition in commit https://github.com/geerlingguy/ansible-role-php-versions/commit/9282f45fded63bcccbcb75d6b6d7dfca3c38c689

ak4t0sh:
@geerlingguy i also facing this issue with debian 9. Tried to install php with opcache manually i confirm debian default priority is 10-opcache :)

MegaphoneJon:
I can confirm that `10-opcache.ini` is also the default in Ubuntu 16.04 with the stock packages (i.e. not Ondrej's PPA).  Merging this locally eliminated the `Cannot load Zend OPcache - it was already loaded` error for me.

MegaphoneJon:
It looks like Travis failed on this because it ran before the Ubuntu 12.04 checks were removed.  If the Travis build is restarted this should pass tests.

geerlingguy:
Doesn't need to pass tests; should once I merge in a second or two!

Thanks for your patience on fixing this everyone!

PR #203, opened 2017-07-03
Add module mcrypt

dthielking:
@geerlingguy I thought about your opinions. I hope this will fit into your php role. I just want to activate mcrypt as php module


geerlingguy:
@dthielking - Since mcrypt is not a requirement for many PHP applications, and also since it's particular configuration seems to require a bit of extra code (which I'd rather not maintain in this role), my best suggestion would be to split out mcrypt extension installation and configuration into a separate role. For a simple example, see how I built the `geerlingguy.php-redis`, `geerlingguy.php-mysql`, etc. roles.

zavan:
In case anyone bumps into this and wants to install and enable mcrypt on PHP 7.2, this works for me:

```yml
- name: Setup PHP mcrypt extension
  become: yes
  block:
    - name: Update PECL channel
      command: pecl channel-update pecl.php.net

    - name: Install PHP mcrypt extension from PECL
      shell: echo "\n\n\n\n\n\n\n\n\n" | pecl install mcrypt

    - name: Add mcrypt to PHP mods-available
      lineinfile:
        line: extension=mcrypt.so
        create: yes
        path: "/etc/php/{{ php_version }}/mods-available/mcrypt.ini"

    - name: Enable mcrypt on PHP
      command: phpenmod mcrypt
```

PR #204, opened 2017-07-17
add include path configuration option

styks1987:


templates/php.ini.j2
```diff
@@ -70,6 +70,7 @@ always_populate_raw_post_data = -1
 ; Paths and Directories ;
 ;;;;;;;;;;;;;;;;;;;;;;;;;
 
+include_path = {{ php_include_path }}
```
geerlingguy: Does this need to be quoted?

styks1987: added quotes

PR #206, opened 2017-07-25
Allow override of html_errors setting.

michaelthieulin:


oxyc:
@geerlingguy maybe worth adding jinja blocks to the role for stuff like this?

geerlingguy:
@oxyc - Copied from #210 —

> Yes, that seems like it would be easier for me to stomach. Once PHP starts breaking compatibility with some options, that's going to make this role waaaay more complicated if we keep adding dozens of new variables.

michaelthieulin:
Hi, what we do with this PR.
What do you mean by adding jinja blocks to the role ?

Thanks

PR #210, opened 2017-07-28
Allows completely overriding the php.ini template

nkakouros:
Removes the `php_use_managed_ini` variable and adds a new one `php_ini_template_file` that allows pointing to a specific ini template file.

This is useful in a scenario that I found myself in. I am responsible for setting up the live environment. However, there is someone else that actually sets up some server flavors locally. I clone their repo and with ansible I put, among many other things, their changes online. Some of the things they change is the php.ini. 

Also, it is nice to provide users with a way to supply their ini file, than simply telling them to find a way to do so.

Finally, there is a similar option for the vhosts file in the ansible-role-apache role. 

I also thought of doing the same with the apc template, but I though it would be better after a discussion of this PR.


oxyc:
In addition to this I think it would be useful to add jinja blocks around groups of variables so people can extend the provided template variables that are configurable. Like [`geerlingguy.nginx`](https://github.com/geerlingguy/ansible-role-nginx#overriding-configuration-templates).

geerlingguy:
@oxyc - Yes, that seems like it would be easier for me to stomach. Once PHP starts breaking compatibility with some options, that's going to make this role waaaay more complicated if we keep adding dozens of new variables.

tterranigma:
Is there interest in getting this merged? It is a pretty minor addition.

Regardless of having blocks in the shipped template or not, I think it is a good idea to allow users to completely ditch the shipped template. In many cases, templating is not needed at all and one only needs to upload a hard-coded file.

nkakouros:
Maintaining my own fork.

nkakouros:
Is there interest in getting this merged?

gnd:
Can  you please finally merge this ? It would be very helpful !

PR #214, opened 2017-09-02
Support opcache.fast_shutdown config param

jamieburchell:
Especially useful in situations where Opcache causes Apache to segfault when this directive is enabled

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -38,6 +38,7 @@ php_opcache_revalidate_path: "0"
 php_opcache_revalidate_freq: "2"
 php_opcache_max_file_size: "0"
 php_opcache_blacklist_filename: ""
+php_opcache_fast_shutdown: "1"
 
 # APCu settings.
 php_enable_apc: true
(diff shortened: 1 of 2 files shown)


FinBoWa:
Suggestions to make this pull request something that  @geerlingguy is more likely to merge:

1. keep the commits to a minumum, those files should be in a one commit
2. use PHP defaults.  http://php.net/manual/en/opcache.configuration.php > "opcache.fast_shutdown" default is 0 not 1.

jamieburchell:
Sorry

1. I edited this in Github, I'm not sure how to do that - was trying to be more helpful than opening an issue
2. The default is 1 when installing the php56u-opcache package hence I thought that was the default.

FinBoWa:
@jamieburchell that was a quick suggestion nothing more :). 

Those changes can't be used without each other so it would be more cleaner to keep them in one commit. Actually not sure that can rebase commits directly on github. 

And for the opcache default: ah.. is that a documentation issue on http://php.net/manual/en/opcache.configuration.php or issue with the php56u-opcache defaults :/

jamieburchell:
Not sure, perhaps it changed between PHP versions, or it's something specific to the php56u package, but a fresh install of php56u-opcache creates a config file with that param set to 1 :/

PR #217, opened 2017-09-11
Allow extra config vars for php-fpm pool

logan2211:
Allows deployers to set extra config lines in the php-fpm pool settings (ie. status/ping URIs)


logan2211:
Any feedback? It is a pretty trivial change that would give folks the opportunity to use this role without forking in situations like #244 and #248.

Lusitaniae:
Please merge

PR #220, opened 2017-10-13
Support Archlinux via a generic setup-package.yml

nogweii:
Not certain about this approach. Could be a bit cleaner I reckon, but this gets it done.


blankoworld:
What about merging it? Or reviewing it?

PR #221, opened 2017-10-17
Make PHP version configurable

nesh-younify:
This will work for Debian/Ubuntu only tho.


geerlingguy:
@nesh-younify - At this time, I would rather people use my [`geerlingguy.php-versions`](https://github.com/geerlingguy/ansible-role-php-versions) role for this functionality, or manage the switch in their playbook via other variables and/or adding a PPA they like. Thank you for the PR, though!

PR #223, opened 2017-11-26
update deprecated include and static usage

rockholla:
Pretty simple, just replace relevant `include:` with `include_tasks:` and remove `static:`


rockholla:
OK, sorry for all the reopening, was seeing intermittent issues in some other travis builds over the last week so thought this might be the same. This one can't be merged until `geerlingguy/docker-fedora24-ansible` is updated to Ansible 2.4 which I think might not happen.  2.4 might only be available to 25+.  Closing this for now.

PR #231, opened 2018-03-08
Adding support for Ondrej's PHP PPA for Ubuntu OS to this playbook.

drivard:
It is possible to install the ppa using the variable

	php_ppa_use: true

It is also now possible to change the PHP version to install under Ubuntu by specifying the variable

	php_version: "7.1"

As left default version to the Debian version which is at the moment 7.0.
I wrote the change in this playbook using the same strategy that has been used in the role `ansible-role-nginx`.


FinBoWa:
Have you featured something that's on https://github.com/geerlingguy/ansible-role-php-versions role already?

drivard:
Didn't knew about that php_versions repo, that should be referenced in the README file of the current repo.

PR #235, opened 2018-04-11
Make precision configurable, add php-cgi config

ajardan:
Since some people need a different value for precision (like us), would be nice to have it configurable.


ajardan:
How does it look ? Do I need to do anything else so this get merged ?

Thanks !

geerlingguy:
Just for more background: http://php.net/manual/en/ini.core.php#ini.precision

> The number of significant digits displayed in floating point numbers. -1 means that an enhanced algorithm for rounding such numbers will be used.

(I didn't know if this affected anything else in particular).

ni-ajardan:
@geerlingguy seems like we forgot to merge these. Should I solve the conflicts, and you merge, or is it already there ?

geerlingguy:
I'd be okay with merging just the precision change for now—not sure exactly what the other change is used for (the php-cgi change), and it seems that could cause more issues than allowing precision to be configured.

ni-ajardan:
@geerlingguy that change was for the situation when someone runs php in cgi mode, those paths were not considered. I believe this is a rare case, since the majority uses fpm, so we can omit that, yes.

geerlingguy:
In that case I'd be happy to merge a precision-config-only change.

geerlingguy:
Thank you!

PR #239, opened 2018-05-14
Parameterize php-fpm log value

jyaworski:
Allow setting alternative paths to `php-fpm`'s `error_log`.


jyaworski:
@geerlingguy bump.

PR #242, opened 2018-05-30
Included missing user/group PHP-FPM configurations in README.

vkononov:


geerlingguy:
Technically those are defined at runtime, so the value varies by OS. Instead of putting apache, can you put something like `[default varies by OS]`?

geerlingguy:
LGTM!

PR #246, opened 2018-07-13
Add php 7.2.7

Yecodeo:
upgrade php from 7.0 to 7.2.7
test under ubuntu 16.04 work fine


geerlingguy:
Unfortunately the defaults need to stay at 7.0 for now, because of the wide variety of OSes this role supports. I would like to update the default source version at some point, but it would be best to do that separately from adjusting any other role parameters.

PR #248, opened 2018-07-25
Added listen variables to php-fpm www.conf

jserpapinto:
Listen variables were needed to correctly set permissions and ownership of php-fpm socket.
This is how I solved it, also added them to README and to default variables in order to pass the build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ If you're using Apache, you can easily get it configured to work with PHP-FPM us
 
 Specific settings inside the default `www.conf` PHP-FPM pool. If you'd like to manage additional settings, you can do so either by replacing the file with your own template or using `lineinfile` like this role does inside `tasks/configure-fpm.yml`.
 
-Listen owner, group and mode are set as default. Do override for your usecase.
+Listen owner, group and mode are set as default. Do override them for your usecase.
 
 ### php.ini settings
 
--- a/README.md
+++ b/README.md
@@ -64,8 +64,13 @@ If you're using Apache, you can easily get it configured to work with PHP-FPM us
     php_fpm_pm_start_servers: 5
     php_fpm_pm_min_spare_servers: 5
     php_fpm_pm_max_spare_servers: 5
+    php_fpm_listen_owner: "www-data"
+    php_fpm_listen_group: "www-data"
+    php_fpm_listen_mode: "0660"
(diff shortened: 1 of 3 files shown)


michaljbarczewski:
@geerlingguy Any chance this gets merged?

johnyserpa:
@geerlingguy could you merge this?

biztems:
@geerlingguy +1 here too, we need this in our workflow. Pretty please?

johnyserpa:
@biztems you can clone my fork, but ideally this would be merged..

biztems:
> @biztems you can clone my fork, but ideally this would be merged..

Yes, that's what I've done - but I'd prefer to use Ansible Galaxy and requirements.yml if possible 👍

devantoine:
I was about to make the same PR because if `php_fpm_listen` is configurable we need to configure the owner, group and permissions.

I see two issues with your PR : 

1) Default values for "owner" and "group" should respectively be set to `php_fpm_pool_user` and `php_fpm_pool_group` and not hardcoded as you did because it would only work under Debian based OS, RHEL based OS use "apache" and not "www-data".

2) Default permission should be set to "0600".

johnyserpa:
@devantoine Added your suggestions. Thanks a lot!

devantoine:
@johnyserpa Have you tested it?

Because I did exactly what you did regarding the lineinfile task and was getting a syntax error when running the playbook.

I had to replace double quotes by single quotes but I have no clue why. Maybe someone could enlighten me on this.

Not working:
`- regexp: "^;?listen\.owner.?=?.*$"`

Working:
`- regexp: '^;?listen\.owner.?=?.*$'`

And did you check that an a Debian OS these three values were indeed commented out?

johnyserpa:
@devantoine it is only tested here: https://regex101.com/r/FQPZa1/1

Sorry, I'll dig into this later.

johnyserpa:
@devantoine I've fixed it. Double quotes were throwing error of "unknown unescaped character", but single quotes fixed it. I've tested it locally and it worked.

devantoine:
Thanks for your work @johnyserpa! 
@geerlingguy I think it's ready to be merged.

devantoine:
@geerlingguy This really should be merged 🙏

PR #251, opened 2018-08-24
fixing building from source on centos

bhunt89:
fix for:
fails to run when installing from source:
on centos 7.5/7.4/7.2/6.9 :
failed: [phptest] (item=libmcrypt-devel) => {"changed": false, "item": "libmcrypt-devel", "msg": "No package matching 'libmcrypt-devel' found available, installed or updated", "rc": 126, "results": ["No package matching 'libmcrypt-devel' found available, installed or updated"]}

fatal: [phptest]: FAILED! => {"changed": false, "msg": "Failed to find required executable git in paths: /sbin:/bin:/usr/sbin:/usr/bin:/usr/local/sbin"}

on centos 6.9:
failed: [phpmemcachedtest] (item=mariadb-devel) => {"changed": false, "item": "mariadb-devel", "msg": "No package matching 'mariadb-devel' found available, installed or updated", "rc": 126, "results": ["No package matching 'mariadb-devel' found available, installed or updated"]}

### DIFF
--- a/tasks/install-from-source.yml
+++ b/tasks/install-from-source.yml
@@ -1,7 +1,41 @@
 ---
-- name: Ensure dependencies for building from source are installed (RedHat).
+- name: Ensure dependencies for building from source are installed (CentOS).
+  package:
+    name: epel-release
+    state: present
+  when: ansible_distribution == "CentOS"
+  
+- name: Ensure dependencies for building from source are installed (RedHat < 7).
   package: "name={{ item }} state=present"
   with_items:
+    - git
+    - autoconf
+    - automake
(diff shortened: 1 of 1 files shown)


geerlingguy:
If you need epel, I recommend adding the `geerlingguy.epel` role to install and configure that repository.

bhunt89:
shouldn't it be added in the requirements then?
what about the mariadb/mysql on centos 6.9

geerlingguy:
@bhunt89 - It is only required in some circumstances (and not at all on non-RHEL platforms).

PR #266, opened 2019-05-17
Remove settings which are absent in php 7.0+

guessit-dev:
https://www.php.net/manual/en/ini.core.php#ini.asp-tags
https://www.php.net/manual/en/ini.core.php#ini.always-populate-raw-post-data
https://www.php.net/manual/en/session.configuration.php#ini.session.bug-compat-42
https://www.php.net/manual/en/session.configuration.php#ini.session.bug-compat-warn

Removing them since this role/php only supports [7.1+](https://php.net/supported-versions.php)

### DIFF
--- a/templates/php.ini.j2
+++ b/templates/php.ini.j2
@@ -6,7 +6,6 @@
 
 engine = On
 short_open_tag = {{ php_short_open_tag }}
-asp_tags = Off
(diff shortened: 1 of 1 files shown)


geerlingguy:
LGTM.

PR #269, opened 2019-06-19
Set php-fpm socket owner and group

### LINKED ISSUE #166: php-fpm socket user and group on centos 7 (nginx)
In my playbook using 
>php_fpm_listen: "/dev/shm/php.socket"
php_fpm_pool_user: nginx
php_fpm_pool_group: nginx

when I restart php-fpm the socket is recreated with owner and group "root",  so nginx stop working and returns a 502 error.

I resolve this by setting:
>listen.owner = nginx
listen.group = nginx
listen.mode = 0660

in /etc/php-fpm.d/www.conf.

porcospino:
Fixes #166

### DIFF
--- a/templates/www.conf.j2
+++ b/templates/www.conf.j2
@@ -4,6 +4,9 @@ listen.allowed_clients = 127.0.0.1
 user = {{ php_fpm_pool_user }}
 group = {{ php_fpm_pool_group }}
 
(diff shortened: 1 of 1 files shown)


JasperAlgra:
@geerlingguy  Could this PR please be merged? 
It would be very handy instead of altering the config after installing the role.

geerlingguy:
👍

PR #277, opened 2019-09-11
Php_packages_extra fix

### LINKED ISSUE #276: Combining php_packages and php_packages_extra with set_fact causes unexpected issues
I ran into a very frustrating and very confusing issue. The packages in `php_packages_extra` were not getting installed when I was passing in `php_packages` to this role. I need to directly pass `php_packages` in to the role since my organization is setting up multiple PHP versions on some webservers, so I generate `php_packages` by prepending `php<version>-` to each package in another list.

This bug happens because a variable that is passed into a role has a higher precendence than a fact. When this role `set_fact`s `php_packages`, that fact's value will never be used because ansible will always take the role variable. This is fine and dandy usually, but this role uses `set_fact` to combine `php_packages` with `php_packages_extra`. The combined list never gets referred to and the packages in `php_packages_extra` will never install. 

This issue would be fixed by removing the combination `set_fact` in `main.yml` and adding an `apt` or `yum` that operates on `php_packages_extra`. Or, simply concatenate the lists with `{{ php_packages + php_packages_extra }}` in the `apt`/`yum` instances.

I got my setup to work by using `set_fact` on `php_packages` before including this role, but it was very frustrating and required me to go through documentation and read the source for this role to understand why my configuration wasn't getting applied as expected.

kindlehl:
This fixes #276

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -3,6 +3,9 @@
 # for RHEL/CentOS.
 php_enablerepo: ""
 
+# No extra packages are installed by default
+php_packages_extra: []
+
 # Default PHP version to install on Debian-based OSes.
 php_default_version_debian: "7.0"
 
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -8,11 +8,6 @@
     php_packages: "{{ __php_packages | list }}"
   when: php_packages is not defined
 
-- name: Define extra php_packages.
-  set_fact:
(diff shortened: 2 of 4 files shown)


kindlehl:
@geerlingguy Any way to get this bad boy merged?

geerlingguy:
I like it.

PR #278, opened 2019-09-18
Allow configuring PHP error_log

paste:
Allow specifying `error_log` file in PHP.ini and configuring the owner/group/permissions of the file. An empty file is created if one doesn't exist. This should be mostly harmless, since it defaults to not being set, as it is currently. 

I'm not sure how common it is to configure `error_log`, or if most setups just use the PHP-FPM/Apache log. I prefer the cleaner PHP error log during development.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,13 +96,17 @@ By default, all the extra defaults below are applied through the php.ini include
     php_error_reporting: "E_ALL & ~E_DEPRECATED & ~E_STRICT"
     php_display_errors: "Off"
     php_display_startup_errors: "On"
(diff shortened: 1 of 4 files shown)


psihius:
Could we get this one merged? This is very useful since if you get an error before you set your PHP error handler, the error is not recorded at all now (or you need to manually edit your php.ini on the server).

jamieburchell:
This would be really useful when PHP is used on the CLI outside of PHP-FPM (e.g. cron jobs). At the moment, no PHP errors can be logged.

PR #280, opened 2019-10-15
Use |bool filter on opcache variable to avoid deprecation warning

jeffsheltren:
When running the role, I'm getting a deprecation warning:

```
TASK [geerlingguy.php : Ensure OpCache config file is present.] ****************************************************************************************
[DEPRECATION WARNING]: evaluating 1 as a bare variable, this behaviour will go away and you might need to add |bool to the expression in the future.
Also see CONDITIONAL_BARE_VARS configuration toggle.. This feature will be removed in version 2.12. Deprecation warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
```

This fixes by using the `bool` filter to convert the variable to a boolean for conditional checks.

### DIFF
--- a/tasks/configure-opcache.yml
+++ b/tasks/configure-opcache.yml
@@ -25,7 +25,7 @@
     force: true
     mode: 0644
   with_items: "{{ php_extension_conf_paths }}"
-  when: php_opcache_enable
+  when: php_opcache_enable|bool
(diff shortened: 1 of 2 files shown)


geerlingguy:
Thanks! Can't believe I didn't notice this sooner.

PR #289, opened 2020-02-24
Removing php-imap from RedHat vars

tobiasehlert:
php-imap is not available anymore.. it was epel7 and will not be included in rh8


geerlingguy:
I'm okay with this change since it does break some tests for RHEL 8.

PR #295, opened 2020-03-13
Allow php_opcache to have any number of extra options

laurentdavid:
Aim of this patch is to allow other configuration values to be added to opcache without having to change the base php role values (https://www.php.net/manual/fr/opcache.configuration.php)

* Use a dictionary instead of a fixed set of variables
* Add the default php_opcache_defaults values

Ensure we have a default value for php_opcache dictionary


laurentdavid:
Reopening after fixing build checks.

PR #297, opened 2020-03-27
Access log and status endpoint

adrianlzt:


shakalandy:
Would be great to get this one merged.
Any help needed on that @geerlingguy?

templates/php.ini.j2
```diff
@@ -64,6 +64,11 @@ auto_append_file =
 
 default_mimetype = "text/html"
 
+{% if php_always_populate_raw_post_data is defined %}
+always_populate_raw_post_data = {{ php_always_populate_raw_post_data }}
```
geerlingguy, line 68: This seems to have been put in as well but is not documented anywhere and is not in the defaults.

defaults/main.yml
```diff
@@ -30,6 +30,10 @@ php_fpm_pm_start_servers: 5
 php_fpm_pm_min_spare_servers: 5
 php_fpm_pm_max_spare_servers: 5
 
+php_fpm_access_log: /var/log/php-fpm/access.log
```
geerlingguy, line 33: All the options would need to be documented in the README as well.

Also, for `php_fpm_access_format`, is that the default access format?

templates/php.ini.j2
shakalandy: https://www.php.net/manual/en/ini.core.php - removed in PHP 7.0.0? Maybe remove, here?

defaults/main.yml
adrianlzt: ``access.log`` by default is not defined. How we can handle that?

geerlingguy: That's a good question...

I think at some point, it's easier to just template the php-fpm configuration from a .j2 file in your playbook instead of trying to define things for any of the possible options :/

But access log would be nice to be able to configure. I just like to stick to PHP's defaults if possible.

adrianlzt: I have modified php_fpm_access_format with the default value.
But what's not default is having a log file

PR #298, opened 2020-04-01
Make serialize_precision a parameter

rvdbogerd:
According to the php documentation, the default version for php should be "-1". Before php7.1 it was 17, so I parameterized it, and made the default "-1"

see https://www.php.net/manual/en/ini.list.php

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -76,6 +76,7 @@ php_output_buffering: "4096"
 php_short_open_tag: "Off"
 php_disable_functions: []
 php_precision: 14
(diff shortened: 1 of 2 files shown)


bjdelange:
Good catch !
This is important because this setting will mess up json_encode in php 7.1+

https://stackoverflow.com/questions/42981409/php7-1-json-encode-float-issue

rvdbogerd:
Hi @geerlingguy could you please have a look at this change? I had to fork the repo for this 1 line change :)

PR #300, opened 2020-04-10
Added possibility to choose if php-fpm is restarted or reloaded.

rsicart:
In production systems, it is sometimes better to reload FPM instead of
restarting. This commit allows users to manage restart or reload of php-fpm
service, but keeps restarting by default.

Co-authored-by: Julien LE SAUX <julien.le.saux@fr.clara.net>

### DIFF
--- a/handlers/main.yml
+++ b/handlers/main.yml
@@ -9,7 +9,7 @@
 - name: restart php-fpm
   service:
     name: "{{ php_fpm_daemon }}"
-    state: restarted
+    state: "{{ php_fpm_handler_state }}"
   when:
     - php_enable_php_fpm
     - php_fpm_state == 'started'
--- a/README.md
+++ b/README.md
@@ -67,6 +67,10 @@ If you're using Apache, you can easily get it configured to work with PHP-FPM us
 
 Control over the fpm daemon's state; set these to `stopped` and `false` if you want FPM to be installed and configured, but not running (e.g. when installing in a container).
 
+    php_fpm_handler_state: restarted
+
+The handler restarts PHP-FPM by default. Setting the value to `reloaded` will reload the service, intead of restarting it.
+
     php_fpm_listen: "127.0.0.1:9000"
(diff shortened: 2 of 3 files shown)


ncharlot:
Was looking for this feature.

rsicart:
@geerlingguy do you have some kind of roadmap to share about this role?
I ask to have more visibility about PR reviews and the future of this repo.
Thanks!

handlers/main.yml
```diff
@@ -3,7 +3,7 @@
   service:
     name: "{{ php_webserver_daemon }}"
     state: restarted
-  notify: restart php-fpm
+  notify: "{{ php_fpm_handler_action }} php-fpm"
   when: php_enable_webserver
 
 - name: restart php-fpm
```
geerlingguy, line 9: I'd rather just set the `state` on this task to a variable, instead of adding conditionals inside `notify` on all the tasks and also adding a new handler. Then the user can choose whether the state should be `restarted` or `reloaded` when the `restart php-fpm` handler is notified.

This new variable needs to also be documented in the README.

I would call the variable `php_fpm_handler_state` and default it to `restarted`.

ncharlot: So we'll need to `notify: restart php-fpm` when just reload is wanted.
I think it should be confusing when reading the tasks.

ncharlot: Why having 2 handlers should be worst?

rsicart: In the beginning I made just a variable, but just after I thought the same than @ncharlot : 1 restart handler which doesn't restart but reloads seems confusing to me.
That said, and you being owner of that repo, I can rewrite the branch to do what you asked, my main purpose is to get that feature merged into master :)

rsicart: Variable `php_fpm_handler_state` it is! :)

geerlingguy:
Thanks! This is available in 4.3.0 now.

rsicart:
Happy to contribute! :)

PR #302, opened 2020-06-05
Configure extra fpm pools

rsicart:
Allow to configure extra fpm pools with a generic template.

You can find an example of php_fpm_extra_pools usage in default vars.
pool_name and pool_listen are mandatory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,14 +71,24 @@ Control over the fpm daemon's state; set these to `stopped` and `false` if you w
 
 The handler restarts PHP-FPM by default. Setting the value to `reloaded` will reload the service, intead of restarting it.
 
-    php_fpm_listen: "127.0.0.1:9000"
-    php_fpm_listen_allowed_clients: "127.0.0.1"
-    php_fpm_pm_max_children: 50
-    php_fpm_pm_start_servers: 5
-    php_fpm_pm_min_spare_servers: 5
-    php_fpm_pm_max_spare_servers: 5
+    php_fpm_pools
 
-Specific settings inside the default `www.conf` PHP-FPM pool. If you'd like to manage additional settings, you can do so either by replacing the file with your own template or using `lineinfile` like this role does inside `tasks/configure-fpm.yml`.
+List of PHP-FPM pool to create. By default, www pool is created.
+
+    pool_name: www
+    pool_template: www.conf.j2
+    pool_listen: "127.0.0.1:9000"
+    pool_listen_allowed_clients: "127.0.0.1"
+    pool_pm: dynamic
+    pool_pm_max_children: 50
+    pool_pm_start_servers: 5
+    pool_pm_min_spare_servers: 5
+    pool_pm_max_spare_servers: 5
+    pool_pm_max_requests: 500
+
+To setup a new pool, add an item to php_fpm_pools list.
+
+Specific settings inside the default `www.conf.j2` PHP-FPM pool. If you'd like to manage additional settings, you can do so either by replacing the file with your own template using `pool_template`.
 
 ### php.ini settings
 
--- a/README.md
+++ b/README.md
@@ -71,22 +71,20 @@ Control over the fpm daemon's state; set these to `stopped` and `false` if you w
 
 The handler restarts PHP-FPM by default. Setting the value to `reloaded` will reload the service, intead of restarting it.
 
-    php_fpm_pools
 
-List of PHP-FPM pool to create. By default, www pool is created.
+    php_fpm_pools:
+      - pool_name: www
+        pool_template: www.conf.j2
+        pool_listen: "127.0.0.1:9000"
+        pool_listen_allowed_clients: "127.0.0.1"
+        pool_pm: dynamic
+        pool_pm_max_children: 50
+        pool_pm_start_servers: 5
+        pool_pm_min_spare_servers: 5
+        pool_pm_max_spare_servers: 5
+        pool_pm_max_requests: 500
 
-    pool_name: www
-    pool_template: www.conf.j2
-    pool_listen: "127.0.0.1:9000"
-    pool_listen_allowed_clients: "127.0.0.1"
-    pool_pm: dynamic
-    pool_pm_max_children: 50
-    pool_pm_start_servers: 5
-    pool_pm_min_spare_servers: 5
-    pool_pm_max_spare_servers: 5
-    pool_pm_max_requests: 500
-
-To setup a new pool, add an item to php_fpm_pools list.
+List of PHP-FPM pool to create. By default, www pool is created. To setup a new pool, add an item to php_fpm_pools list.
 
 Specific settings inside the default `www.conf.j2` PHP-FPM pool. If you'd like to manage additional settings, you can do so either by replacing the file with your own template using `pool_template`.
 
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -37,16 +37,6 @@ php_fpm_pools:
     pool_pm_min_spare_servers: 5
     pool_pm_max_spare_servers: 5
     pool_pm_max_requests: 500
-#  - pool_name: pool
-#    pool_template: www.conf.j2
-#    pool_listen: 127.0.0.1:9001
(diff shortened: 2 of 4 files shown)


rsicart:
What do you think about that PR @geerlingguy ?

ncharlot:
Hi @rsicart ! We follow definitely the same features :)

I think it would be better to use a common var for all pools with default to the `www` basic one as it exists now. And we could share the same template tooother params).

The default vars should be like:

    php_fpm_pools:
      - name: www
        listen: 127.0.0.1
        user: www-data
        ...

And maybe a `template` attribute as @geerlingguy do with Nginx role:

    php_fpm_pools:
      - name: www
        template: "{{ php_fpm_pool_template }}"
        listen: 127.0.0.1
        user: www-data
        ...

rsicart:
Hahah I see @ncharlot ! I have often cases where I need several pools :)
And well, I implemented fpm extra pools separated from default www pool because, in case that festure is not interesting for @geerlingguy , I can easily keep my fork up to date with this repo without conflicts. It’s less intrusive.
Nevertheless, I agree with you: fpm pool setup should be the same logic for 1 and for N. And of course, allowing the user to choose the template!

ncharlot:
Yes for performance and readability purpose we for use to split apps parts to differents pools.

rsicart:
I rewrote the PR to use a single list of pools, as @ncharlot proposed: `php_fpm_pools`
You can find an example of php_fpm_pools usage in default vars, and the documentation in `README.md`

ncharlot:
Ahah good job @rsicart!

rsicart:
Hi there @geerlingguy ,
do you have some feedback about this PR please?
Thanks in advance

rsicart:
bump

JKetelaar:
Would love this see this feature introduced as well!

JKetelaar:
Could you review this @geerlingguy?

EdinTC:
This feature would be an great addition!

ncharlot:
@geerlingguy I'm sorry to insist but adding such a common feature in the main repo seems to be nice rather in a fork.

geerlingguy:
One thing I definitely would not like to do is break backwards compatibility. Probably 99.9% (maybe 100%) of this roles _current_ users are happy with the single defined www pool, as the use case for this role is most frequently "one server, one webserver, one site" (or maybe a few / dozen / more, but still behind one pool).

This PR breaks that backwards compatibility. At minimum I'd rather have this be opt-in functionality, and have it be able to use the normal / standard way it uses now, and you could add additional pools (or named pools and disable the standard pool the role currently sips with) if desired through the new variable.

rsicart:
Or we could release a major version, for example 5.0.0, including a changelog if needed :)

geerlingguy:
@rsicart - The problem is, this is a feature that I would guess relatively few people (of all this role's users) use, and I don't want to change the default way the entire role works to cater to the minority of the role's users.

More specifically, I have never used more than the single default node pool, and that's why this PR has sat for a long time. I would rather not merge something into a role I use a _lot_ that will require me to change dozens of playbooks just to keep the role working ;)

So it's less of an ask, and more of a demand.

rsicart:
No worries, I completely understand your point of view. I just hope you do not refuse only because you'll have to update your excellent book Ansible for Devops ! ;)

@all What do you think about something more like this commit, when this branch was less intrusive (before PR rewrite on June 26th)? https://github.com/geerlingguy/ansible-role-php/commit/e732bafeac010874003a7512458b4f46f5b815a5
Would that be acceptable for all of us?

ncharlot:
To preserve BC, maybe we should just keep the current vars for values of the 1st pool:

    php_fpm_listen: "127.0.0.1:9000"
    php_fpm_listen_allowed_clients: "127.0.0.1"
    php_fpm_pm_max_children: 50
    php_fpm_pm_start_servers: 5
    php_fpm_pm_min_spare_servers: 5
    php_fpm_pm_max_spare_servers: 5

    # PHP-FPM pool configuration.
    php_fpm_pools:
     - pool_name: www
        pool_template: www.conf.j2
        pool_listen: "{{ php_fpm_listen }}"
        pool_listen_allowed_clients: "{{ php_fpm_listen_allowed_clients }}"
        pool_pm: dynamic
        pool_pm_max_children: {{ php_fpm_pm_max_children }}
        pool_pm_start_servers: {{ php_fpm_pm_start_servers }}
        pool_pm_min_spare_servers: {{ php_fpm_pm_min_spare_servers }}
        pool_pm_max_spare_servers: {{ php_fpm_pm_max_spare_servers }}
        pool_pm_max_requests: 500

rsicart:
Good point! That seems a good compromise.

mcdir:
MR https://github.com/geerlingguy/ansible-role-php/pull/321 with a good compromise ;)

README.md
```diff
@@ -71,14 +71,24 @@ Control over the fpm daemon's state; set these to `stopped` and `false` if you w
...
 
-    php_fpm_listen: "127.0.0.1:9000"
-    php_fpm_listen_allowed_clients: "127.0.0.1"
-    php_fpm_pm_max_children: 50
-    php_fpm_pm_start_servers: 5
-    php_fpm_pm_min_spare_servers: 5
-    php_fpm_pm_max_spare_servers: 5
-
-Specific settings inside the default `www.conf` PHP-FPM pool. If you'd like to manage additional settings, you can do so either by replacing the file with your own template or using `lineinfile` like this role does inside `tasks/configure-fpm.yml`.
+    php_fpm_pools
```
geerlingguy, line 74: It looks like the formatting here got screwed up a little bit. This should be displaying the same as it is in defaults.yml—right now it looks like `php_fpm_pools` is just a bare variable... and the pool below is just floating in space on its own.

Maybe print the entire `php_fpm_pools` var just like it is in defaults, then drop the documentation line about what this is below the code block.

defaults/main.yml
```diff
@@ -24,12 +24,29 @@ php_enable_php_fpm: false
...
+  - pool_name: www
+    pool_template: www.conf.j2
+    pool_listen: "127.0.0.1:9000"
+    pool_listen_allowed_clients: "127.0.0.1"
+    pool_pm: dynamic
+    pool_pm_max_children: 50
+    pool_pm_start_servers: 5
+    pool_pm_min_spare_servers: 5
+    pool_pm_max_spare_servers: 5
+    pool_pm_max_requests: 500
```
geerlingguy, line 39: This is a new option—do you know if `500` is the default for max_requests?

geerlingguy (changes requested):

defaults/main.yml
rsicart: Default value is 0, which means "forever" according to documentation.

Here is a snippet from a fresh install on debian buster, from www.conf file:

```
; The number of requests each child process should execute before respawning.
; This can be useful to work around memory leaks in 3rd party libraries. For
; endless request processing specify '0'. Equivalent to PHP_FCGI_MAX_REQUESTS.
; Default Value: 0
;pm.max_requests = 500
```

Personally I'd set this value > 0, for example 2000.

geerlingguy (approved):

geerlingguy:
This has been merged and will be in the 4.6.0 release of the role. Thanks for your patience!

rsicart:
Thanks to you and all people who collaborated along this PR :)

ncharlot:
Congrats @rsicart!

geerlingguy:
@tjoelsson - Shoot! I didn't even think about that when I merged... can you open a new issue so we can figure out a better way of handling backwards compatibility for those upgrading this role?

tjoelsson:
@geerlingguy Sure, here is the issue https://github.com/geerlingguy/ansible-role-php/issues/329

geerlingguy:
@tjoelsson - Thanks; if nobody can help with getting the backwards compatibility sorted soon, I'm willing to revert this PR and release a new version of the role, then delay adding it back in until a compatibility layer is added back in.

rsicart:
What about what @ncharlot proposed in this comment  ?
https://github.com/geerlingguy/ansible-role-php/pull/302#issuecomment-741609370

PR #303, opened 2020-06-15
Add cookie_httponly and cookie_secure to managed php.ini

nemmons:
These are two fairly important security options that should usually be set to true on a production website, so it seems like would make sense to add these to the managed php.ini file for convenience. Of course, I default to false here to maintain backwards compatibility.


Lusitaniae:
Need this too

Lusitaniae:
Still relevant

nemmons:
Still relevant

PR #310, opened 2020-08-24
Allow the php version to be overridden.

dlundgren:
I found out I was unable to specify 7.3 on an Ubuntu 18 installation, and it stemmed from a clean up that enforced the default version being configured on Debian based systems.

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -6,10 +6,11 @@
     - files:
         - "{{ ansible_distribution }}-{{ ansible_distribution_major_version }}.yml"
       skip: true
+
 - name: Set the default PHP version for Debian-based OSes.
   set_fact:
     php_default_version_debian: "{{ __php_default_version_debian }}"
-  when: ansible_os_family == 'Debian'
+  when: php_default_version_debian is not defined and ansible_os_family == 'Debian'
 
 - name: Include OS-specific variables.
   include_vars: "{{ ansible_os_family }}.yml"


JKetelaar:
+1

shakalandy:
Hey @geerlingguy - any chance to push that one?

in our condition - having host_vars overwriting defaults this ended in PHP 7.4 being installed (desired state), but PHP 7.2 being configured - where PHP 7.4 should have been configured.
With the fix from this PR - this works.

piotrkochan:
I had to clone and symlink role to my playbook as it is very important for our usecase. @geerlingguy, come on :)

twoequaldots:
+1.  This patch worked for me to use php7.4 on Debian 10.  If it helps at all, I've used Ansible 2.9.6 - 2.10.2 for my tests and they all passed. :)

geerlingguy:
`php_default_version_debian` should still be able to override this, is it not working?

geerlingguy:
Ah... looking into this more, this was introduced in https://github.com/geerlingguy/ansible-role-php/commit/1496cf8f92de830842b8be2e1e14981bf1b3e1ae — and that commit helped make the role work better across many versions of distros, but made it so you couldn't override it anymore (which this PR fixes).

But that whitespace issue is from https://github.com/geerlingguy/ansible-role-php/commit/9751de2e2a688d3a521af652f4cff4c255b9734f, which is weird, I'm not sure how I let that sneak in!

twoequaldots:
Thanks for looking into this and approving the merge, much appreciated =]

shakalandy:
Thanks also for creating an release for this @geerlingguy  🥇

PR #318, opened 2020-11-16
Manage php-fpm.conf

dlundgren:
Right now php-fpm.conf is only created when installing from source.

This will enable the ability to customize the php-fpm.conf global options


viscat:
It would be nice to have this feature, can be this be merged to master please?

dlundgren:
@geerlingguy Would you be able to review the PR's that have gone stale?

mortenbirkelund:
@geerlingguy Any thoughts about this PR? Would be usefull.

PR #319, opened 2020-12-15
Add new php.ini options

rook59:
Add more php.ini configuration options:
`session.name`, `html_errors` and `error_log`


jamieburchell:
This would be really useful when PHP is used on the CLI outside of PHP-FPM (e.g. cron jobs). At the moment, no PHP errors can be logged.

PR #321, opened 2021-01-31
Configure extra fpm pools and add php_fpm_clear_env parameter

mcdir:
Configure extra fpm pools and add php_fpm_clear_env parameter


templates/www.conf.j2
```diff
@@ -1,15 +1,19 @@
...
+; {{ ansible_managed }}
+
+[{{ item.pool_name | mandatory }}]
+listen = {{ item.pool_listen | mandatory }}
+listen.allowed_clients = {{ item.pool_listen_allowed_clients | default('127.0.0.1', true) }}
 user = {{ php_fpm_pool_user }}
 group = {{ php_fpm_pool_group }}
 
 listen.owner = {{ php_fpm_pool_user }}
 listen.group = {{ php_fpm_pool_group }}
```
pfrenssen, line 10: I think it would be a good idea to make these individually configurable. This makes it possible to create a shared server setup in which each user has a separate pool, and runs the PHP process under their own user, while still making sure the webserver can access the sockets.

For example:
```
user = pieter
group = pieter

listen.owner = www-data
listen.group = www-data
listen.mode = 660
```

shakalandy: There is a related PR in #302 with overlapping features regarding multiple FPM pools.
Would be really cool to have this as an optional configuration - could you also please take a look @geerlingguy and advise on how to proceed? Thanks!

geerlingguy:
Closing to consolidate discussion into #302.

PR #322, opened 2021-02-06
Add possibility to define the mysqli default port

morpheus-87:
This PR adds the possibility to define the mysqli default port via a new variable.


morpheus-87:
Any news on this, @geerlingguy ?

PR #325, opened 2021-03-17
Add php_fpm_memory_limit variable to www.conf.j2

Maksold:
Sometimes I need to have a different memory_limit for CLI and PHP-FPM. 

So I added a new variable php_fpm_memory_limit, which allows to define memory_limit for PHP-FPM.


shakalandy:
Clearly makes sense, thanks for the PR.
Usually cronjobs might use lot more memory as FPM should be allowed to.

PR #326, opened 2021-03-25
Feature: www pool & php 8 jit variables

wells:
- Update `www.conf` template to use all current variables
- Update `configure-fpm` task to overwrite `www.conf` with template every time
- Add `php_fpm_pm` and `php_fpm_pm_max_requests` variables for `www.conf` template
- Add `php_opcache_jit` and `php_opcache_jit_buffer_size` variables for `opcache.ini` template

### DIFF
--- a/templates/www.conf.j2
+++ b/templates/www.conf.j2
@@ -1,15 +1,15 @@
 [www]
-listen = 127.0.0.1:9000
-listen.allowed_clients = 127.0.0.1
+listen = {{ php_fpm_listen }}
+listen.allowed_clients = {{ php_fpm_listen_allowed_clients }}
 user = {{ php_fpm_pool_user }}
 group = {{ php_fpm_pool_group }}
 
 listen.owner = {{ php_fpm_pool_user }}
 listen.group = {{ php_fpm_pool_group }}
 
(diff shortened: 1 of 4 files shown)


shakalandy:
Please have a look here @geerlingguy as it enhances features regarding PHP 8.x - thanks!

templates/www.conf.j2
```diff
@@ -1,15 +1,15 @@
...
 listen.owner = {{ php_fpm_pool_user }}
 listen.group = {{ php_fpm_pool_group }}
 
-pm = dynamic
-pm.max_children = 50
-pm.start_servers = 5
-pm.min_spare_servers = 5
-pm.max_spare_servers = 5
-pm.max_requests = 500
+pm = {{ php_fpm_pm }}
```
geerlingguy, line 10: I think we should limit this PR's scope. #302 already looks at rearranging the way pools work. This PR should focus on just the JIT variables, IMO.

geerlingguy (changes requested):

geerlingguy:
After merging #302 this PR is a bit messy — if there are still other changes can you consider submitting a new one?

PR #334, opened 2021-05-31
Use comment filter on ansible_managed var in www.conf

dannycobussen:
When setting `ansible_managed` to a multiline value, the default www-pool file becomes invalid.
I've added the `comment` filter with a custom decoration.

### DIFF
--- a/templates/www.conf.j2
+++ b/templates/www.conf.j2
@@ -1,4 +1,4 @@
-; {{ ansible_managed }}
+{{ ansible_managed | comment(decoration='; ') }}
 
 [{{ item.pool_name | mandatory }}]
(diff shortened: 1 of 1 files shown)


gotmax23:
I see no reason not to merge this change.

dannycobussen:
@geerlingguy do you want to merge this?

geerlingguy:
@dannycobussen - I've never set ansible_managed to anything more than a small bit of marker text, so never ran into a use case where it would span multiple lines. But I see no big deal with adding the extra jinja here to ensure multiline comments will work. Still seems strange to me, but I know some places are more particular about file headers.

PR #335, opened 2021-07-01
Added support for Ubuntu 21.04 hirsute

basictheprogram:


gotmax23:
@geerlingguy,

I know you only test/target LTS versions of Ubuntu; however, would you please consider merging this change to provide basic, best effort compatibility with Ubuntu Hirsute?

Thanks,
Maxwell

geerlingguy:
Unfortunately, I do not want to add support for any non-LTS releases in my roles, since I don't deploy them on any of my servers (and use Docker for instances where I'd be running PHP locally for development on systems that might use non-LTS releases).

PR #342, opened 2021-08-24
Add support for Debian 11

VincentLanglet:
Hi @geerlingguy, I recently tried to use this library with Debian 11 and it wasn't working without this variable.
Defining the variable myself solve the build, so I suppose there is nothing more to do.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,8 @@ jobs:
             playbook: converge.yml
           - distro: ubuntu1804
             playbook: converge.yml
+          - distro: debian11
+            playbook: converge.yml
           - distro: debian10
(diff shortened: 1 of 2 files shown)


geerlingguy:
Great! Can you also add Debian 11 to the tests in the .github/workflows/ci.yml list (somewhere around halfway down the file, just add a line for `debian11` above the existing `debian10` line). That way CI will also be able to verify the fix.

VincentLanglet:
> Great! Can you also add Debian 11 to the tests in the .github/workflows/ci.yml list (somewhere around halfway down the file, just add a line for `debian11` above the existing `debian10` line). That way CI will also be able to verify the fix.

Sure, done

PR #346, opened 2021-09-28
Allow php_opcache to have any number of extra options

laurentdavid:
When configuring opcache we might need to have additional settings that are not necessary in the original list of default options (see https://www.php.net/manual/fr/opcache.configuration.php)

* Use a dictionary instead of a fixed set of opcache options
* Add the default php_opcache_defaults values
* Ensure we have a default value for php_opcache dictionary


piotrkochan:
I'm just putting/overriding extra opcache settings in conf.d directory, that way I don't need to change anything in the role.

PR #351, opened 2021-12-10
Add pm.status_path to settings and template

### LINKED ISSUE #350: Unable to set `pm.status_path`
Title says it all, really; the current template and vars don't allow you to set `pm.status_path` for a pool, unless I'm missing something obvious — which is far from impossible.

Is this an actual decision or just something that simply never came up and that you're willing to receive a PR for?

doenietzomoeilijk:
Adding a PR to #350 just in case. ;)

This should allow you to set a `pm.status_path` for an FPM pool. Leaving it empty should, [per the documentation](https://www.php.net/manual/en/install.fpm.configuration.php#pm.status-path), disable the feature for that pool.

Open to comments and suggestions, I'm not too experienced with Ansible templating.

Closes #350

### DIFF
--- a/defaults/main.yml
+++ b/defaults/main.yml
@@ -31,6 +31,7 @@ php_fpm_pm_start_servers: 5
 php_fpm_pm_min_spare_servers: 5
 php_fpm_pm_max_spare_servers: 5
 php_fpm_pm_max_requests: 0
+php_fpm_pm_status_path: ""
 
 # PHP-FPM pool configuration.
 php_fpm_pools:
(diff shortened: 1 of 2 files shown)


shakalandy:
Thanks @doenietzomoeilijk!
Sounds like a reasonable add-on, what do you think @geerlingguy ?

doenietzomoeilijk:
I really hope this could still be picked up at some point, or if it's not up to snuff, explicitly closed.

shakalandy:
Hey @geerlingguy - would be really cool to have this one merged - thx!

geerlingguy:
Please see my comment on https://github.com/geerlingguy/ansible-role-php/issues/350#issuecomment-1073316880 — I think that would be more flexible for these types of settings that aren't commonly used by most PHP installations.

PR #352, opened 2021-12-13
Fix variable name for pm.max_requests.

doenietzomoeilijk:
While combing through the options, I encountered this default with a "wrong" name. This would result in pools not getting the for-all-pools value set for `pm.max_requests`, instead falling back to the default 500.


doenietzomoeilijk:
I really hope this could still be picked up at some point, or if it's not up to snuff, explicitly closed.

racke:
Bump

PR #353, opened 2021-12-21
Added possibility to choose if web server is restarted or reloaded

KuiKui:
In production systems, it is sometimes better to reload web server instead of restarting it. This commit allows users to manage restart or reload of web server service, but keeps restarting it by default.

Thanks to @rsicart for #300


spocky:
To add more context, our web app (web site files, but also lower level components such as php or elasticsearch) can be updated by the click of a button. 
Once playbooks are updated and start playing, we faced what first appeared as a random crash with no log whatsoever.
The bottom of it is that if php (or a few other components ftm) gets updated, an ansible handler is automatically set to restart apache. This in turns kills our update script before it has a chance to finish.
Setting this value to 'reload' allows the process to finish gracefully which totally fixes this behavior.

Jstham:
It is indeed a problem as our automation would fall down every now and then.

PR #355, opened 2022-01-05
OPCache: set JIT buffer size to enable JIT compiler

simonspa:
PHP8's JIT is only enabled when an explicit JIT buffer size larger than 0 is set.
This MR exposes the relevant variable through this ansible role.


simonspa:
I would argue this is a simple-enough and relevant addition.

simonspa:
Still applies cleanly to master and is an easy fix to enable JIT

simonspa:
Still relevant I believe. @geerlingguy care to have a look and merge?

Issue #358, opened 2022-01-28
php8.0-json no installation candidate


dgibbs64:
Re opening based off this issue that went stale https://github.com/geerlingguy/ansible-role-php/issues/323

php-json package is no longer necessary for PHP 8.0: reference
Because php-json is included in default __php_packages list for Debian-based distributions, installation fails:

E: Package 'php8.0-json' has no installation candidate
Solution: install php{{ php_default_version_debian }}-json package only when php_default_version_debian less than 8.0

Failure output
```
TASK [geerlingguy.php : Ensure PHP packages are installed.] ********************
fatal: [forest-tools.com]: FAILED! => {
    "cache_update_time": 1643369106,
    "cache_updated": false,
    "changed": false,
    "rc": 100
}

STDOUT:

Reading package lists...
Building dependency tree...
Reading state information...
Package php8.0-json is a virtual package provided by:
  php8.0-phpdbg 8.0.14-1+ubuntu20.04.1+deb.sury.org+1
  php8.0-fpm 8.0.14-1+ubuntu20.04.1+deb.sury.org+1
  php8.0-cli 8.0.14-1+ubuntu20.04.1+deb.sury.org+1
  php8.0-cgi 8.0.14-1+ubuntu20.04.1+deb.sury.org+1
  libphp8.0-embed 8.0.14-1+ubuntu20.04.1+deb.sury.org+1
  libapache2-mod-php8.0 8.0.14-1+ubuntu20.04.1+deb.sury.org+1

STDERR:

E: Package 'php8.0-json' has no installation candidate

MSG:

'/usr/bin/apt-get -y -o "Dpkg::Options::=--force-confdef" -o "Dpkg::Options::=--force-confold"       install 'php8.0-common' 'php8.0-cli' 'php8.0-dev' 'php8.0-fpm' 'libpcre3-dev' 'php8.0-gd' 'php8.0-curl' 'php8.0-imap' 'php8.0-json' 'php8.0-opcache' 'php8.0-xml' 'php8.0-mbstring' 'php-sqlite3' 'php-apcu' 'php8.0-curl' 'php8.0-xml' 'php8.0-zip' -o APT::Install-Recommends=no' failed: E: Package 'php8.0-json' has no installation candidate
```

dgibbs64:
Its worth noting that I didnt notice this issue with 8.1 so I might but be an issue with 8.0

dgibbs64:
[...]

PR #369, opened 2022-05-17
Add variables for Ubuntu 22

jschlier:
Hey there geerlingguy,

we have started setting up Ubuntu 22.04 servers and noticed that this role breaks because of a missing variable file.
I added it and set the variable accordingly.

### DIFF
--- a/vars/Ubuntu-22.yml
+++ b/vars/Ubuntu-22.yml
@@ -1,0 +1,2 @@
+---
+__php_default_version_debian: "8.1"


dgibbs64:
Please also note this issue. I am unsure the best way to deal with the requirement to remove php8.1-json package as it does not exist in php 8 https://github.com/geerlingguy/ansible-role-php/issues/358

dgibbs64:
bump

dgibbs64:
bump

PR #388, opened 2022-11-11
Add php_fpm_pool_listen_user and php_fpm_pool_listen_group

duviful:
Add the ability to specify listen.owner and listen.group, which can be distinct from user and group.


dlundgren:
Ideally this should be using `{{ item.pool_user | default(php_fpm_pool_user) }}` and `{{ item.pool_group | default(php_fpm_pool_group) }}`, so that we can specify these individually per pool, as it is right now, I have to manage the fpm pools independently because I can't specify the user.

shakalandy:
can we get this one merged, please @geerlingguy?

PR #391, opened 2023-01-05
replace with_flattened with loop and flatten filter

gk-codes:
`with_flattened` is replaced by `loop` and the `flatten filter`.

https://docs.ansible.com/ansible/latest/playbook_guide/playbooks_loops.html#with-flattened


gk-codes:
Bugs in suggested code.

PR #394, opened 2023-03-17
Fix variable name for pm.max_requests.

blizzz:
reopening https://github.com/geerlingguy/ansible-role-php/pull/352

I face the same problem, and the same change fixes it.

@geerlingguy to your attention please


blizzz:
what can be done to drive this forward?

PR #367, opened 2022-03-30
Allow extra settings per pool, fix #350

steinbrueckri:

PR #395, opened 2023-03-17
add option to set pm.status_path

blizzz:
Makes it possible to define a status_path. The default value is empty, and in that case the setting will not be written to the config file, i.e. no change to default behaviour.

for your consideration @geerlingguy


blizzz:
what can be done to drive this forward?

blizzz:
rebased and resolved conflict

steinbrueckri:
See also https://github.com/geerlingguy/ansible-role-php/pull/367

PR #406, opened 2023-05-05
add support for Ubuntu 22.04 (jammy)

j-koehler:
...and remove trusty from meta file since it is already no longer supported

/cc @geerlingguy


j-koehler:
friendly bump

PR #415, opened 2024-03-26
Can select PHP version on RedHat.

### LINKED ISSUE #414: Select PHP version on RedHat
# Description

If I'm not mistaken, one cannot select the PHP version to install on RedHat.

# Why is it important?

This is critically important to me because I am developing this role that installs a website that requires PHP >= 7.4 (namely dokuwiki).

Now, in my opinion, my `ansible-role-dokuwiki` is totally responsible for installing the PHP version it requires.

Right now, this is done by selecting the appropriate module before running the php role:

`ansible-role-dokuwiki/tasks/main.yml`:
```yaml
- name: Selects PHP:7.4 (RedHat)
  command: "dnf module enable -y php:7.4"
  when: ansible_os_family == "RedHat"

- name: Install php
  include_role: 
    name: geerlingguy.php
```

Also, my role defines `geerlingguy.php` as a dependency in `meta/main.yml`.

# Problem

As ansible runs the dependencies before the current role tasks, it results in a bug due to RedHat's (rightful) refusal to select a different PHP version after installing certain packages.

# Proposed solution

1. Defining `php_default_version_redhat` in `defaults/main.yml`
2. Include this in `tasks/main.yml`

```yaml
- name: Set the default PHP version for RedHat-based OSes.
  command: "dnf module enable -y php:{{ php_default_version_redhat }}"  
  when: php_default_version_redhat is defined and ansible_os_family == 'RedHat'
```

# Workaround

The workaround would be to select the appropriate PHP module in the playbook that runs `ansible-role-dokuwiki`:

`server-setup.yml`:
```yaml
---
- hosts: all
  pre_tasks:
   - name: Selects PHP:7.4 (RedHat)
     command: "dnf module enable -y php:7.4"
     when: ansible_os_family == "RedHat"

  roles:
    - ansible-role-dokuwiki
```

This is not acceptable for me, as, again, to me, `ansible-role-dokuwiki` is suppose to install the adequate version of PHP, depending on the dokuwiki version the user is ultimately installing.

benoistlaurent:
Fixes #414

### DIFF
--- a/tasks/setup-RedHat.yml
+++ b/tasks/setup-RedHat.yml
@@ -1,4 +1,7 @@
 ---
+- name: Set the default PHP version for RedHat-based OSes.
+  command: "dnf module enable -y php:{{ __php_default_version_redhat }}"
+
 - name: Ensure PHP packages are installed.
   package:
     name: "{{ php_packages + php_packages_extra }}"
--- a/tasks/setup-RedHat.yml
+++ b/tasks/setup-RedHat.yml
@@ -1,6 +1,6 @@
 ---
 - name: Set the default PHP version for RedHat-based OSes.
-  command: "dnf module enable -y php:{{ __php_default_version_redhat }}"
+  command: "dnf module enable -y php:{{ php_default_version_redhat }}"
 
 - name: Ensure PHP packages are installed.
   package:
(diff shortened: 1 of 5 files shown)


tasks/setup-RedHat.yml
```diff
@@ -1,4 +1,7 @@
 ---
+- name: Set the default PHP version for RedHat-based OSes.
+  command: "dnf module enable -y php:{{ php_default_version_redhat }}"
```
eb4x, line 3: this is not idempotent. try

```
package:
  name: "@php:{{ php_default_version_redhat }}/common"
```
if you want to install from appstreams.

PR #418, opened 2024-05-15
feat: allow the configuration of any opcache variable

jpiron:
This removes the need to update the role to support new opcache variables.


shakalandy:
Cool, this would make my PR #416 obsolete, but keep the default, right?

jpiron:
Yes, except for the `php_realpath_cache_ttl` variable. \
And indeed it keeps all the defaults as long as you don't define the `php_opcache_*` variables.\
I updated the README to make it more clear.

shakalandy:
@geerlingguy could we get this one merged? I would adapt #416 to include `php_realpath_cache_ttl` only - thanks!

PR #425, opened 2024-08-05
Allow user to override php.ini.j2 template file

dw-ec:
Other templates such as www.conf.j2 can already be overridden by the user. Allow the php.ini.j2 template file to be overridden too.


olherasymchuk:
Thank you @dw-ec ,

With your changes I using playbook with something like this: `php_ini_template: "{{ playbook_dir }}/templates/php.ini.j2"` - and it works as expected.

Sad to see this PR was not merged.

Issue #429, opened 2024-10-22
Make the role compatible with Rocky linux


SonOfWinter:
The role does not currently work with systems from the Rocky Linux family.

racke:
Which version of ansible-core do you use?

SonOfWinter:
ansible 2.10.6
python 3.9.18

Is it possible that the version I'm using doesn't know Rocky linux?

racke:
Yes, that is really likely.  Ansible is EOL since 23 May 2022.

SonOfWinter:
I updated it and now everything works correctly.
Thank you for your help

PR #430, opened 2024-10-22
Add Rocky Linux configuration

SonOfWinter:
Refer to #429

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -25,6 +25,11 @@
     __php_packages: "{{ __php_packages + [__php_json_package_debian] }}"
   when: __php_json_package_debian is defined and __php_json_package_debian not in __php_packages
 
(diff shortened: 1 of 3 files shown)


cinnion:
I would love to know what the exact issue was, because I recently used the role to install not only one version of PHP, but PHP 7.4 through 8.3 on a RL 9.4 system with zero issues as a part of a server to upgrade the PHP version to the latest version. The ansible server was also running the latest stable version of Ansible on RL 9.4, though most of the work was actually done on a CentOS 7 box running Ansible 2.9.

See [cinnion/ansible-role-medusa](https://github.com/cinnion/ansible-role-medusa) for an example.

SonOfWinter:
I was using an old version of ansible that didn't recognize RL. 
I updated it and now everything works correctly.
My PR is now obsolete and I've closed it.

PR #435, opened 2025-02-19
Update www.conf.j2

c-perronnet:
adding the option to configure the values for
request_terminate_timeout 
request_terminate_timeout_track_finished

in the pool template


c-perronnet:
not ready

PR #438, opened 2025-08-28
feat: allow to not managed opcache/apcu and fpm if not needed

Poil:


Poil:
Not stale

deaniliev:
This will also allow installing of php 8.5 as there is no opcache module yet. I want to disable it completely as the role fails with opcache enabled.

PR #439, opened 2025-10-21
Add Debian 13 support

clement-gouin:
Debian 13 default php version is 8.4


mikevo:
Did this work? I needed to remove the `php-8.4-imap` package and replace `libpcre3-dev` with `libpcre2-dev` as well. See #441

PR #440, opened 2025-10-30
Add Debian-13 support

tobias-richter:
* Debian 13 php version is 8.4
* php 8.4. does no longer ship with imap

### DIFF
--- a/vars/Debian-13.yml
+++ b/vars/Debian-13.yml
@@ -1,0 +1,15 @@
+---
+__php_default_version_debian: "8.4"
+__php_packages:
+  - php{{ php_default_version_debian }}-common
+  - php{{ php_default_version_debian }}-cli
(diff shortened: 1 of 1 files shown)


mikevo:
Did this work? I needed to replace `libpcre3-dev` with `libpcre2-dev` as well. See #441

tobias-richter:
@mikevo I had only ran them against a already existing installation. Just checked with a vanilla Debian 13 instance and I also ran into the issue.
```
TASK [geerlingguy.php : Ensure PHP packages are installed.] ************************************************************
fatal: [debian-13-testing]: FAILED! => {"changed": false, "msg": "No package matching 'libpcre3-dev' is available"}
```
So your PR https://github.com/geerlingguy/ansible-role-php/pull/441/ looks better :) Closing mine

Issue #444, opened 2026-03-27
imap removed from PHP 8.5 causes role to fail


bigcalm:
Role version: 6.0.0

Using the ansible role on Ubuntu 26.04 (yes, I'm testing the beta) fails when it attempts to install `php8.5-imap`:

```
TASK [geerlingguy.php : Ensure PHP packages are installed.] ************************************************************************************************
[ERROR]: Task failed: Module failed: No package matching 'php8.5-imap' is available
Origin: /home/iain/kubuntu-2604-test/ansible/roles/geerlingguy.php/tasks/setup-Debian.yml:5:3

3   apt: update_cache=yes cache_valid_time=86400
4
5 - name: Ensure PHP packages are installed.
    ^ column 3

fatal: [iain-precisision5530]: FAILED! => {"changed": false, "msg": "No package matching 'php8.5-imap' is available"}
```

https://www.php.net/manual/en/book.imap.php states that

> This extension is DEPRECATED and UNBUNDLED as of PHP 8.4.0.

PR #402, opened 2023-04-19
Only install the JSON extension on Debian for PHP 7.4 and below

pfrenssen:
Starting with PHP 8.0 the JSON extension is part of the base package and cannot be installed separately.

Fixes #358

PR #445, opened 2026-03-29
Bugfix/php8.4 and up imap removal

bigcalm:
The IMAP extension has been removed from PHP from version 8.4.

Install candidates have since been removed from Debian unstable.

See: https://tracker.debian.org/news/1672837/removed-3103-1-from-unstable/

2nd commit - based upon PR #402 - installs the extension if the PHP version is lower than 8.4.

Fixes issue #444

### DIFF
--- a/tasks/main.yml
+++ b/tasks/main.yml
@@ -25,6 +25,16 @@
     __php_packages: "{{ __php_packages + [__php_json_package_debian] }}"
   when: __php_json_package_debian is defined and __php_json_package_debian not in __php_packages
 
+- name: Define the name of the IMAP extension package on Debian for PHP <8.4.
+  set_fact:
+    __php_imap_package_debian: "{{ 'php' + php_default_version_debian + '-imap' }}"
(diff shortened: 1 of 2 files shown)


pfrenssen:
Thanks very much for working on this! I think the proposed solution certainly achieves the goal but is maybe not the best solution to facilitate future maintenance of the role.

If we now start managing version-specific extension overrides in `main.yml` tasks this will set us on a path towards a proliferation of similar tasks when more and more extension changes are introduced in new PHP versions.

In the role we handle overrides in the distro-specific vars files. I do agree that this is not the best possible solution (after all, the IMAP extension removal is specific to PHP 8.4, not necessarily to a certain version of a distro), but it will make it more easily manageable. It also points users to the fact they can take control of the variables for their own machines.

bigcalm:
It's been a while since I worked on this, but I think I was mirroring how the role catered for when JSON was absorbed into the core of PHP.

Happy to see the role working with Ubuntu 26.04 with your PR instead.

PR #449, opened 2026-04-10
Ubuntu 25: PHP 8.4

belal-i:


belal-i:
Hi, this should make the role work for Ubuntu 25. I did a quick test, and everything worked. LMK if further changes are required :-)

